PR #8, opened 2018-07-31
Conv2d

koenvanderveen:


src/tensorflow_encrypted/layer/core.py
```diff
@@ -1,13 +1,20 @@
...
+from ..protocol.pond import Pond
 
+DEFAULT_PROTOCOL = Pond
+
+# TODO
+# split backward function in compute_gradient and compute_backpropagated_error ?
 
 class Layer:
-    pass
+    prot = DEFAULT_PROTOCOL
```
mortendahl: how about a `set_protocol` and `get_protocol` methods instead that update and read a global variable? was thinking something along the `with Pond(..) as prot:` pattern, where the `__enter__` method of a protocol calls `set_protocol`.

src/tensorflow_encrypted/layer/core.py
```diff
@@ -42,19 +48,92 @@ def backward(self, d_y, learning_rate):
...
+        self.layer_output = None
 
     def initialize(self, input_shape, initializer=None):
         pass
 
     def forward(self, x):
-        #TODO[koen]: same story
-        y = prot.sigmoid(x)
-        self.layer_ouput = y
+        y = self.prot.sigmoid(x)
```
mortendahl: any opinion on what's better: `prot.sigmoid(x)` or `x.sigmoid()`?

src/tensorflow_encrypted/protocol/pond.py
```diff
@@ -1263,24 +1293,107 @@ def _dot_masked_masked(prot, x, y):
...
     z = PondPrivateTensor(z0, z1)
     z = prot.truncate(z)
     return z
 
+#
+# Conv helpers
+#
+# TODO[koen] create operations for all possible combinations
+
+def _conv2d_int100(x, y, strides, padding="SAME"):
```
mortendahl: seems to make sense to move this to the backing tensor..?

src/tensorflow_encrypted/tests/conv2d_test.py
```diff
@@ -0,0 +1,66 @@
+import tensorflow as tf
+import numpy as np
+import unittest
+from tensorflow_encrypted.protocol import Pond
+from tensorflow_encrypted.protocol import Server
+from tensorflow_encrypted.layer import Conv2D, set_protocol
+from tensorflow_encrypted.config import session
+
+
+class TestConv2D(unittest.TestCase):
```
mortendahl: brilliant!

src/tensorflow_encrypted/layer/core.py
koenvanderveen: okay, set_protocol and get_protocol functions of which class then?

src/tensorflow_encrypted/protocol/pond.py
koenvanderveen: yes definitely

mortendahl: done

src/tensorflow_encrypted/layer/core.py
mortendahl: done

Issue #14, opened 2018-08-13
Explicit typing


mortendahl:
Currently there's a lot of type assertions in the code to fail fast in case the wrong type is being passed to a method. Most if not all of these could be replaced by [type annotations and the `mypy` type checker](http://mypy.readthedocs.io/en/stable/index.html).

bendecoste:
I can take a look at this

PR #30, opened 2018-08-13
Types

bendecoste:
This is generally what I was thinking for types.  Like @morgangiraud pointed out if I add types to every single function at once it will cause lots of conflicts so maybe we slowly roll over to types as you work around the code?

related #14


tensorflow_encrypted/protocol/pond.py
```diff
@@ -97,7 +97,7 @@ def define_public_variable(self, initial_value, apply_encoding=True, name=None):
         _initializers.append(x.initializer)
         return x
 
-    def define_private_variable(self, initial_value, apply_encoding=True, name=None):
+    def define_private_variable(self, initial_value, apply_encoding=True, name=None) -> PondPrivateTensor:
```
mortendahl: I think you want this to be `PondPrivateVariable`, otherwise you might run into problems when trying to assign to it later!

PR #31, opened 2018-08-13
Package

### LINKED ISSUE #20: Python package
Make the library available as an easy to use package.

jvmancuso:
Closes #20.

This PR makes tf-encrypted into a Python package.  It refactors some of the directory structure to follow Python best practices, checks for dependencies (e.g. numpy and tensorflow), and describes installation instructions toward the end of the README.

In the future, we'll want to think about handling Tensorflow installation.  Unfortunately, there's no easy way to do this.  There are two separate PyPI packages for TF (`tensorflow` and `tensorflow-gpu`) depending on GPU usage.  Ideally we'd install `tensorflow-gpu` when we detect a CUDA-compatible device, but that's apparently not foolproof as it can mess up when using Docker.  For now, I just added them both as extra dependencies, which means the user has to bring their own version of Tensorflow that satisfies our requirements (specified in the value for the `extra_requires` argument to the `setup` function in `setup.py`).


setup.py
```diff
@@ -0,0 +1,14 @@
+from setuptools import setup
+
+setup(
+    name = 'tf-enrypted',
```
mortendahl: typo

PR #35, opened 2018-08-14
stub instructions

bendecoste:
See README diff for examples on how this actually gets used.


bendecoste:
> Looks good, stubs/numpy/init.pyi is empty. It should probably be removed?

Left it in there so it is more obvious how to setup a new library whenever the time comes .. I don't think it's a big deal to sit there.

justin1121:
If `MYPYPATH` is set to the stubs folder any code that uses numpy, mypy will complain about it not being able to find that attribute in the numpy module.

image

justin1121:
Could just do a test package instead or something!

bendecoste:
> If MYPYPATH is set to the stubs folder any code that uses numpy, mypy will complain about it not being able to find that attribute in the numpy module.

pretty sure that is because the ini file is empty.  We should probably use the official one anyway! You're saying rename `numpy` in stubs to something like `example`?

justin1121:
Yeah! I'd be fine with that or just putting the official one in our repo some how.

bendecoste:
> Yeah! I'd be fine with that or just putting the official one in our repo some how.

Thought about doing that but i think we'd have to use submodules which i'm not a huge fan of

Issue #26, opened 2018-08-13
Load a pretrained model for prediction


ianlivingstone:
Details to come

justin1121:
Right now I'm working through a vgg16 model from keras exported as a tensorflow model where the weights and everything are frozen directly into the protobuf graph. This contains a lot of ops we'll need to begin so its a good place to start experimenting with how these ops will work in MPC land. 

Right now I'm just reading the graph and redoing the ops directly in normal tensorflow. Once I get this all sorted I'll look into making this work in mpc land!

PR #37, opened 2018-08-14
convert tensorflow protobuf to MPC graph

justin1121:
Closes https://github.com/mortendahl/tf-encrypted/issues/26

Here's some code that converts a protobuf tensorflow graph into code that just builds a mpc graph.


tensorflow_encrypted/convert/register.py
```diff
@@ -0,0 +1,112 @@
+import tensorflow as tf
+
+from typing import Any, Dict, List
+
+
+def register() -> Dict[str, Any]:
+    reg = {}
```
mortendahl: you could write this like 
```return {
  'Placeholder': placeholder,
  'Const': constant,
  ...
}```

File: examples/convert.py [hunk 2]
```diff
@@ -0,0 +1,53 @@
...
+
+def export_vgg16():
+    vgg16 = tf.keras.applications.VGG16()
+
+    sess = K.backend.get_session()
+
+    pred = [None]
+    pred_node_names = [""]
+    pred_node_names[0] = "output"
+    pred[0] = tf.identity(vgg16.outputs[0], name=pred_node_names[0])
```
morgangiraud: Why do you define `pred = [None]` and then reassign it directly `pred[0] = ...`?
Same question for `pred_node_names `

tensorflow_encrypted/convert/register.py
```diff
@@ -0,0 +1,112 @@
...
+        'Reshape': reshape,
+        'MatMul': matmul,
+        'Softmax': softmax,
+    }
+
+    return reg
+
+
+def placeholder(node: Any, inputs: List[str], output_lookup: Dict[str, Any]) -> Any:
+    return tf.placeholder(node.attr["dtype"].type,
```
morgangiraud: I guess the goal here is to replace those with `tfe` functions?

examples/convert.py
justin1121: Not reason really. This code was borrowed from elsewhere and I haven't make it nice nice yet.

tensorflow_encrypted/convert/register.py
justin1121: You got it!

justin1121:
Yeah not sure how important those tests would be right now. I've started getting rid of the 1-to-1 mapping stuff and the final MPC graph will be wildly different from the input graph.

mortendahl:
let me take a look later today!

justin1121:
Sounds good! Sorry it got so big 😦

PR #44, opened 2018-08-15
run matmul for ci

bendecoste:
this isn't asserting anything but for where we are we can run this for now to ensure
that the code actually runs for CI (ie no import errors etc)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,27 @@
+absl-py==0.3.0
+asn1crypto==0.24.0
+astor==0.7.1
+certifi==2018.8.13
+cffi==1.11.5
+colorama==0.3.7
+cryptography==2.3
+gast==0.2.0
+grpcio==1.14.1
+idna==2.7
+jmespath==0.9.3
+jwt==0.5.4
+Markdown==2.6.11
+mypy==0.620
+numpy==1.14.5
+protobuf==3.6.1
+pyasn1==0.4.2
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -0,0 +1,27 @@
...
+mypy==0.620
+numpy==1.14.5
+protobuf==3.6.1
+pyasn1==0.4.2
+pycparser==2.18
+rsa==3.4.2
+s3transfer==0.1.13
+six==1.11.0
+tensorboard==1.10.0
+tensorflow==1.10.0
```
jvmancuso: concerned that this will break when users want to use tensorflow-gpu instead of tensorflow.  in the `setup.py` I'd made both of those optional and assumed that the user had installed one of them manually with pip before installing tf-encrypted.  I'd suggest we keep the requirements.txt file similar and try to manually specify our tensorflow version for circleci

requirements.txt
```diff
@@ -0,0 +1,27 @@
...
+termcolor==1.1.0
+-e git+git@github.com:mortendahl/tf-encrypted.git@a880bc038f0368a646f7ba28b3b31a4a5c0fc477#egg=tf_encrypted
```
jvmancuso: why the specific commit hash here?

bendecoste: i generated this with `pip freeze` .. can probably just remove this

Issue #3, opened 2018-06-05
First steps for neural networks


mortendahl:
Implement Keras-like layers and models for easily express (sequential) neural networks along the lines of what's done in [pond](https://github.com/mortendahl/privateml/blob/master/image-analysis/pond/nn.py).

Add additional matching example.

bendecoste:
Most of these are now implemented.  I think we can close this issue and open new issues for more layers as they are needed.

PR #45, opened 2018-08-15
Add Relu Layer

yanndupis:
Hey @jvmancuso ,

Here is the relu layer with only forward for now #3 

Thank you!


tensorflow_encrypted/layers/activation.py
```diff
@@ -17,3 +17,20 @@ def backward(self, d_y, *args):
...
+        pass
+
+    def forward(self, x):
+        y = self.prot.relu(x)
+        self.layer_output = y
+        return y
+
+    # TODO Approximate Relu derivate to implement backward
+    def backward(self, d_y, *args):
+        pass
```
jvmancuso: I'd suggest a `raise NotImplementedError` here

yanndupis: good point!

yanndupis:
The comments should be addressed now. Thanks

PR #47, opened 2018-08-16
Inputs, part 1

mortendahl:


tensorflow_encrypted/protocol/pond.py
```diff
@@ -120,6 +116,51 @@ def define_private_variable(self, initial_value, apply_encoding=True, name=None)
...
+        with tf.name_scope('output{}'.format('-'+name if name else '')):
+
+            with tf.device(receiver.player.device_name):
+
+                v: BackingTensor = _reconstruct(x0, x1)
+                # v = _decode(v, apply_scaling) # TODO[Morten] re-enable
+                op = receiver.receive_output(v.backing[0])
+
+                # wrap in tf.group to prevent sending back any tensors (which might hence be leaked)
+                op = tf.group(op)
```
morgangiraud: Interesting, why does `tf.group` avoid sending back tensors?

tensorflow_encrypted/protocol/pond.py
```diff
@@ -756,16 +767,51 @@ def __init__(self, prot, variable0, variable1):
...
+    encoded = (rationals * scaling_factor).astype(int).astype(object)
+    return BackingTensor.from_native(encoded)
+
+def _decode_to_numpy(elements, scaling_factor: int):
     map_negative_range = np.vectorize(lambda element: element if element <= M/2 else element - M)
-    return map_negative_range(elements).astype(float) / (2**precision)
+    return map_negative_range(elements).astype(float) / scaling_factor
+
+def _encode_from_tftensor(rationals: tf.Tensor, scaling_factor: int) -> BackingTensor:
+    encoded = tf.cast(rationals * scaling_factor, BackingTensor.int_type)
```
morgangiraud: How does this work?
Is there still the risk that we might overflow the `int_type`?

tensorflow_encrypted/__init__.py
```diff
@@ -1,19 +1,17 @@
...
-    'protocol'
-]
+# __all__ = [
+#     'InputProvider',
+#     'LocalConfig',
+#     'RemoteConfig',
+#     'estimator',
+#     'layer',
+#     'protocol'
+# ]
```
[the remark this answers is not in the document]
mortendahl: just easier to play around with for now :) bad practice?

tensorflow_encrypted/protocol/pond.py
mortendahl: I can't find the link I initially used, but you essentially want something that returns an tf.Operation instead of tf.Tensor -- [tf.group](https://www.tensorflow.org/api_docs/python/tf/group) does that

tensorflow_encrypted/protocol/pond.py
mortendahl: not when the plaintext values are small enough (say integral part is < 2^16). the problem is in decoding since the current method for that uses large numbers

tensorflow_encrypted/protocol/pond.py
morgangiraud: In this case, maybe put this line inside `receive_output `?

mortendahl: as you also suggested, wanted to enforce this one and hence put it in the protocol

PR #56, opened 2018-08-16
Add Batchnorm Layer

yanndupis:
Hey

This PR should add the [batchnorm](https://www.tensorflow.org/api_docs/python/tf/nn/batch_normalization) layer #3 .

Thanks


tensorflow_encrypted/layers/batchnorm.py
```diff
@@ -0,0 +1,52 @@
+import numpy as np
+import math
+from . import core
+
+
+class Batchnorm(core.Layer):
+    def __init__(self, mean, variance, scale, offset, variance_epsilone=1e-8):
```
bendecoste: can you add type annotations where possible?

tensorflow_encrypted/layers/batchnorm.py
```diff
@@ -0,0 +1,52 @@
...
+        # Batchnorm after Dense layer
+        if len(input_shape) == 2:
+            N, D = input_shape
+            self.mean = self.mean.reshape(1, N)
+            self.variance = self.variance.reshape(1, N)
+            self.scale = self.scale.reshape(1, N)
+            self.offset = self.offset.reshape(1, N)
+
+        # Batchnorm after Conv2D layer
+        elif len(input_shape) == 4:
```
bendecoste: should there be a `NotImplementedError` if this comes after a conv3d case or other?

PR #60, opened 2018-08-17
strided slice

justin1121:


tensorflow_encrypted/protocol/pond.py
```diff
@@ -434,6 +434,29 @@ def transpose(self, x):
 
         return x_t
 
+    # see https://www.tensorflow.org/api_docs/python/tf/strided_slice for documentation on
+    # the arguments
+    def strided_slice(self, x, *args, **kwargs):
+        node_key = ('strided_slice', x)
+        x_t = _nodes.get(node_key, None)
```
mortendahl: this `_t` naming was for transpose; maybe replace them with `_slice` for clarity?

tensorflow_encrypted/protocol/pond.py
```diff
@@ -1572,10 +1595,72 @@ def _transpose_masked(prot, x_masked):
...
+
+def _strided_slice_masked(prot, x_masked, args, kwargs):
+    assert isinstance(x_masked, PondMaskedTensor)
+
+    a, a0, a1, alpha_on_0, alpha_on_1 = x_masked.unwrapped
+
+    with tf.name_scope('strided_slice'):
+
+        with tf.device(prot.crypto_producer.device_name):
+            a_t = a.strided_slice()
```
mortendahl: you're not passing `args` and `kwargs` in neither of the three applications here

tensorflow_encrypted/protocol/pond.py
```diff
@@ -434,6 +434,29 @@ def transpose(self, x):
...
+    # see https://www.tensorflow.org/api_docs/python/tf/strided_slice for documentation on
+    # the arguments
+    def strided_slice(self, x, *args, **kwargs):
+        node_key = ('strided_slice', x)
+        x_t = _nodes.get(node_key, None)
+
+        if x_t is not None:
+            return x_t
+
+        if isinstance(x, PondPublicTensor):
```
bendecoste: should we use the dispatch pattern from `pond`? maybe not enough types to be worth it

tensorflow_encrypted/protocol/pond.py
justin1121: Code catch 😄

tensorflow_encrypted/protocol/pond.py
justin1121: Other unary ops do it like this.

PR #65, opened 2018-08-17
Add crt_sum

jvmancuso:
Allows for summing over a particular dimension within a single CRT tensor, and extends that functionality to Int100.


tests/convolution_test.py
```diff
@@ -18,14 +18,10 @@ def test_forward(self):
...
-        config = tfe.LocalConfig([
-            'server0',
-            'server1',
-            'crypto_producer'
-        ])
+        config = tfe.LocalConfig(3)
 
         # convolution pond
-        with tfe.protocol.Pond(*config.get_players('server0, server1, crypto_producer')) as prot:
+        with tfe.protocol.Pond(*config.players) as prot:
```
justin1121: Setting up the pond this way doesn't work anymore. Needs to be reverted to how it was.

jvmancuso:
Finished in #91

Issue #22, opened 2018-08-13
Easy deployment to AWS/GCP


mortendahl:
Make it easy to quickly get up and running on AWS or GCP. Could take the form of documentation or code (perhaps along [these lines](https://github.com/mortendahl/privateml/tree/master/tensorflow/spdz/configs/gcp))

ianlivingstone:
Prepackaged configs for kubernetes might be a good way to get this deployed now that they both support k8s as a service.

PR #67, opened 2018-08-20
Add Dockerfile 

ianlivingstone:
This request contains three changes:

- A simple Dockerfile for containerizing tf-encrypted.
- The semver was invalid (should be `-rc`)
- Updated tensorflow requirements to the latest stable version (1.10.0)

Related #22


bin/serve
```diff
@@ -0,0 +1,36 @@
...
+
+config = tfe.RemoteConfig({
+    'server0': server0_address,
+    'server1': server1_address,
+    'crypto_producer': producer_address
+})
+
+server = config.server(name)
+server.start()
+server.join()
```
ianlivingstone: There is no server.stop() exposed in the Python API (it exists in the C API), so I don't believe I can add a signal handler to shut the server down unfortunately.

bin/serve
```diff
@@ -0,0 +1,36 @@
...
+import sys
+import os
+import tensorflow_encrypted as tfe
+
+name = os.environ.get('TFE_NODE_NAME')
+server0_address = os.environ.get('TFE_SERVER0_ADDRESS')
+server1_address = os.environ.get('TFE_SERVER1_ADDRESS')
+producer_address = os.environ.get('TFE_CRYPTO_PRODUCER_ADDRESS')
+
+if name is None:
```
ianlivingstone: It'd be nice if this was refactored a bit to be pushed as much as possible into `tensorflow_encrypted/config.py`.

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM python:3.6.3
```
ianlivingstone: There is probably a much smaller container we can build using alpine or similar

.circleci/config.yml
```diff
@@ -53,6 +55,10 @@ jobs:
             python examples/int100.py
             python examples/inputs.py
             python examples/matmul.py
+      - run:
```
ianlivingstone: We need to use a tool like `make` or similar to wrap up all these different commands.

PR #99, opened 2018-09-07
Improve mypy support

morgangiraud:


Makefile
```diff
@@ -66,6 +67,11 @@ test: lint pythoncheck
 lint: pythoncheck
 	flake8
 
+typecheck: pythoncheck
```
ianlivingstone: We should add this to the `make test` command right?

It should probably also be specified in the `.PHONY`

morgangiraud: Yes for `.PHONY`

Not yet for make test, because it breaks too much for now

morgangiraud: I actually think, someone should focus on mypy for 2 days, fix everything ad add it to `make test` and force other contributors to comply afterward.

ianlivingstone: How about we land this in master and then open another issue for converting the rest of the code base?

morgangiraud: Agree

ianlivingstone: 👍 Let's get the PHONY updated and an issue, then I'll 🍏 :)

PR #109, opened 2018-09-17
native tensor

justin1121:


tensorflow_encrypted/config.py
```diff
@@ -37,7 +37,7 @@ def to_dict(self) -> Dict:
 
 class LocalConfig(Config):
     """
-    Configure tf-encrypted to use threads on the local CPU
+    Configure tf-encrypted to use threads on the local CPU.
```
mortendahl: something is wrong with the sentence now.. :)

tensorflow_encrypted/protocol/pond.py
```diff
@@ -216,8 +216,8 @@ def helper(v: tf.Tensor):
             assert v.shape.is_fully_defined(), "Shape of input '{}' on '{}' is not fully defined".format(
                 name if name else '', provider.player.name)
 
-            v = _encode(v, apply_scaling)
-            x0, x1 = _share(v)
+            var = _encode(v, apply_scaling)
```
mortendahl: `v` was for "value" here, `var` could be misleading

tensorflow_encrypted/protocol/pond.py
```diff
@@ -272,16 +272,16 @@ def helper(x: 'PondPrivateTensor'):
                 if isinstance(xs, PondPrivateTensor):
                     # single input -> single output
                     x = xs
-                    op = receiver.receive_output(helper(x))
+                    t = receiver.receive_output(helper(x))
```
mortendahl: `op` is better: `receive_output` doesn't return a tensor, it returns an operation to call when the output should be pushed to the player

tensorflow_encrypted/tensor/int100.py
```diff
@@ -307,7 +307,7 @@ def from_native(value: np.ndarray) -> 'Int100Constant':
         return Int100Constant(value, None)
 
     @staticmethod
-    def from_int100(value: Int100Tensor) -> 'Int100Constant':
+    def from_backing(value: Int100Tensor) -> 'Int100Constant':
```
mortendahl: `backing` makes less sense here since it expresses what it's used for in Pond more than what it is. I suggest `from_same` instead of `from_backing`

tensorflow_encrypted/tensor/native.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def __mul__(self, other: Union['NativeTensor', int]) -> 'NativeTensor':
+        return self.mul(other)
+
+    def __mod__(self, k: Union['NativeTensor', int]) -> 'NativeTensor':
+        return self.mod(k)
+
+    def add(self, other: Union['NativeTensor', int]) -> 'NativeTensor':
+        x, y = _lift(self, self.modulus), _lift(other, self.modulus)
+        return NativeTensor(x.value + y.value % self.modulus, self.modulus)
```
mortendahl: `%` binds strong than `+`, so you have to do `(x.value + y.value) % self.modulus`; same for `-`

tensorflow_encrypted/tensor/native.py
```diff
@@ -0,0 +1,215 @@
...
+        return NativeTensor(tf.matmul(x.value, y.value) % self.modulus, self.modulus)
+
+    # def im2col(self, h_filter, w_filter, padding, strides) -> 'NativeTensor':
+    #     return _im2col(self, h_filter, w_filter, padding, strides)
+
+    # def conv2d(self, other, strides, padding='SAME') -> 'NativeTensor':
+    #     return _conv2d(self, other, strides, padding)
+
+    def mod(self, k: Union['NativeTensor', int]) -> 'NativeTensor':
+        x, y = _lift(self, self.modulus), _lift(k, self.modulus)
```
mortendahl: suggest you let `k` be an int for consistency (ie don't lift to tensor)

tensorflow_encrypted/io.py
```diff
@@ -21,11 +23,11 @@ def provide_input(self) -> tf.Tensor:
 
 class OutputReceiver(object):
 
-    def __init__(self, player: Player, output_fn: Optional[Callable[[tf.Tensor], tf.Tensor]] = None) -> None:
+    def __init__(self, player: Player, output_fn: OutputFunc = None) -> None:
         self.player = player
         self.output_fn = output_fn
 
-    def receive_output(self, tensor: tf.Tensor) -> tf.Tensor:
+    def receive_output(self, tensor: Union[tf.Tensor, List[tf.Tensor]]) -> tf.Tensor:
```
mortendahl: could you change the return type to `tf.Operation` btw? returning `tf.Tensor` is a mistake (https://github.com/mortendahl/tf-encrypted/blame/master/tensorflow_encrypted/io.py)

tensorflow_encrypted/protocol/pond.py
mortendahl: use `w` or `val` perhaps?

tensorflow_encrypted/protocol/pond.py
mortendahl: see comments below; signature of `receive_output` was changed recently but shouldn't have been

tensorflow_encrypted/protocol/pond.py
justin1121: Yep sure!

tensorflow_encrypted/tensor/native.py
justin1121: Ahh yes, good call!

tensorflow_encrypted/tensor/native.py
```diff
@@ -0,0 +1,215 @@
...
+        x, y = _lift(self, self.modulus), _lift(other, self.modulus)
+        return NativeTensor(tf.matmul(x.value, y.value) % self.modulus, self.modulus)
+
+    # def im2col(self, h_filter, w_filter, padding, strides) -> 'NativeTensor':
+    #     return _im2col(self, h_filter, w_filter, padding, strides)
+
+    # def conv2d(self, other, strides, padding='SAME') -> 'NativeTensor':
+    #     return _conv2d(self, other, strides, padding)
+
+    def mod(self, k: Union['NativeTensor', int]) -> 'NativeTensor':
```
mortendahl: type of `k` just int, otherwise it'll fail :)

PR #111, opened 2018-09-17
First steps for SecureNN

mortendahl:


tensorflow_encrypted/protocol/pond.py
```diff
@@ -20,7 +21,7 @@
 )
 from ..io import InputProvider, OutputReceiver
 from ..player import Player
-from .protocol import Protocol, _global_cache_updators
+from .protocol import Protocol, _global_cache_updators, memoize, _nodes
```
justin1121: should the leading underscores be removed from `_global_cache_updators` and `_nodes` if they're not private?

jvmancuso: I'm in favor

mortendahl: done!

PR #112, opened 2018-09-17
native pond

justin1121:
This PR adds experimental use of the `NativeTensor` in pond so that we can make use of a tensor to add features into securenn. It also implements binarize for the `NativeTensor`.

### DIFF
--- a/examples/securenn-playground.py
+++ b/examples/securenn-playground.py
@@ -7,14 +7,14 @@ config = tfe.LocalConfig([
     'crypto_producer'
 ])
 
-with tfe.protocol.SecureNN(*config.get_players('server0, server1, crypto_producer')) as prot:
+with tfe.protocol.SecureNN(*config.get_players('server0, server1, crypto_producer'), use_native=True) as prot:
 
     a = prot.define_constant(np.array([0, 0, 1, 1]), apply_scaling=False)
     b = prot.define_constant(np.array([0, 1, 0, 1]), apply_scaling=False)
     c = prot.bitwise_or(a, b)
 
-    x = prot.define_constant(np.array([0., 1., 2., 3.]))
-    y = prot.define_constant(np.array([0., 1., 2., 3.]))
+    x = prot.define_constant(np.array([0., 1., 2., 3.]), apply_scaling=False)
+    y = prot.define_constant(np.array([0., 1., 2., 3.]), apply_scaling=False)
     z = (x * c) * y
 
     with config.session() as sess:
--- a/tensorflow_encrypted/protocol/pond.py
+++ b/tensorflow_encrypted/protocol/pond.py
@@ -14,6 +14,8 @@ from ..tensor.int100 import (
     Int100Placeholder as BackingPlaceholder,
     stack,
 )
+
+
 from ..tensor.helpers import (
     log2,
     gcd,
@@ -45,6 +47,24 @@ _initializers: List = list()
 _thismodule = sys.modules[__name__]
 
 
+def import_native() -> None:
+    global BackingTensor
+    global BackingConstant
+    global BackingVariable
+    global BackingPlaceholder
+    global stack
+    from ..tensor.native import (
+        NativeTensor as BackingTensor,
+        NativeConstant as BackingConstant,
+        NativeVariable as BackingVariable,
+        NativePlaceholder as BackingPlaceholder,
+        stack,
+    )
+
+    global M
+    M = 2**16
+
+
 class Pond(Protocol):
 
     def __init__(
@@ -52,12 +72,16 @@ class Pond(Protocol):
         server_0: Player,
         server_1: Player,
         crypto_producer: Player,
-        use_noninteractive_truncation: bool=False
+        use_noninteractive_truncation: bool=False,
+        use_native: bool=False
     ) -> None:
         self.server_0 = server_0
         self.server_1 = server_1
         self.crypto_producer = crypto_producer
         self.use_noninteractive_truncation = use_noninteractive_truncation
+
+        if use_native:
+            import_native()
 
     def define_constant(self, value: np.ndarray, apply_scaling: bool=True, name: Optional[str]=None) -> 'PondConstant':
         assert isinstance(value, (np.ndarray,)), type(value)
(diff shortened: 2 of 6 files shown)


tensorflow_encrypted/tensor/native.py
```diff
@@ -90,6 +95,28 @@ def strided_slice(self, args: Any, kwargs: Any) -> 'NativeTensor':
...
+        if isinstance(self.value, np.ndarray):
+            bitwidths = np.arange(bits, dtype=np.int64)
+            for i in range(len(self.shape)):
+                bitwidths = np.expand_dims(bitwidths, 0)
+            val = np.expand_dims(self.value, -1)
+            val = np.right_shift(val, bitwidths) & 1
+
+            return NativeTensor.from_native(val, p)
+        elif isinstance(self.value, tf.Tensor):
+            bitwidths = np.arange(bits, dtype=np.int64)
```
jvmancuso: should this be the `tf.range`?

tensorflow_encrypted/tensor/native.py
```diff
@@ -90,6 +95,28 @@ def strided_slice(self, args: Any, kwargs: Any) -> 'NativeTensor':
...
+            for i in range(len(self.shape)):
+                bitwidths = np.expand_dims(bitwidths, 0)
```
jvmancuso: should this be `tf.expand_dims`?

tensorflow_encrypted/tensor/native.py
```diff
@@ -90,6 +95,28 @@ def strided_slice(self, args: Any, kwargs: Any) -> 'NativeTensor':
     def reshape(self, axes: Union[tf.Tensor, List[int]]) -> 'NativeTensor':
         return NativeTensor(tf.reshape(self.value, axes), self.modulus)
 
+    def binarize(self) -> 'NativeTensor':
+        """Computes bit decomposition of tensor
+         tensor: ndarray of shape (x0, ..., xn)
+        returns: a binary tensor of shape (x0, ..., xn, bits) equivalent to tensor
+        """
+        if isinstance(self.value, np.ndarray):
```
jvmancuso: Is there a particular use case we have in mind for keeping numpy stuff in here? I feel like the `value` constructor arg could be converted to a tf tensor during construction if it's a numpy array

jvmancuso:
also, can we add some unit tests for NativeTensor in this PR?  something with a bit more coverage than securenn-playground.py would be excellent (in particular, the binarize test from our research repo would be useful here).

examples/securenn-playground.py
```diff
@@ -7,14 +7,14 @@
...
-with tfe.protocol.SecureNN(*config.get_players('server0, server1, crypto_producer')) as prot:
+with tfe.protocol.SecureNN(*config.get_players('server0, server1, crypto_producer'), use_native=True) as prot:
 
     a = prot.define_constant(np.array([0, 0, 1, 1]), apply_scaling=False)
     b = prot.define_constant(np.array([0, 1, 0, 1]), apply_scaling=False)
     c = prot.bitwise_or(a, b)
 
-    x = prot.define_constant(np.array([0., 1., 2., 3.]))
-    y = prot.define_constant(np.array([0., 1., 2., 3.]))
+    x = prot.define_constant(np.array([0., 1., 2., 3.]), apply_scaling=False)
```
mortendahl: scaling should be applied here and below, any reason for changing it?

tensorflow_encrypted/protocol/pond.py
```diff
@@ -45,20 +47,42 @@
 _thismodule = sys.modules[__name__]
 
 
+def import_native() -> None:
```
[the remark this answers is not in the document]
mortendahl: i'm in favour of doing this the right way now tbh since we're soon going to be offering several backing tensors. jump on a call today when you guys are up?

tensorflow_encrypted/protocol/pond.py
```diff
@@ -67,10 +91,10 @@ def define_constant(self, value: np.ndarray, apply_scaling: bool=True, name: Opt
         with tf.name_scope('constant{}'.format('-' + name if name else '')):
 
             with tf.device(self.server_0.device_name):
-                x_on_0 = BackingConstant.from_same(v)
+                x_on_0 = BackingConstant.from_same(v, M)
```
mortendahl: let's find a way to avoid this as it'll be weird otherwise: sometimes `M` is respected by the backing tensor, sometimes not

tensorflow_encrypted/protocol/pond.py
```diff
@@ -1121,9 +1145,10 @@ def _decode(elements: BackingTensor, is_scaled: bool) -> Union[tf.Tensor, np.nda
     with tf.name_scope('decode'):
 
         scaling_factor = 2 ** BITPRECISION_FRACTIONAL if is_scaled else 1
+        b = BOUND if is_scaled else 1
```
mortendahl: why 1 if not scaled?

tensorflow_encrypted/protocol/pond.py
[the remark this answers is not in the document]
justin1121: Yep, have an idea on how we could do this the right way myself. Best to chat and pick the right one!

tensorflow_encrypted/protocol/pond.py
justin1121: Down!

examples/securenn-playground.py
justin1121: I was just trying to get my weird hack to return the right numbers and adding `apply_scaling=False`, same reason I added the bound hack below. Figured its not the right way of doing things.

tensorflow_encrypted/tensor/native.py
justin1121: Yeah was thinking about this a bit and it shouldn't really matter if its a numpy array or a tf tensor because the tf ops take numpy arrays as input.

tensorflow_encrypted/tensor/native.py
justin1121: numpy does work here and below but will change it to tf to be consistent!

justin1121:
Closing this guy for now.

PR #114, opened 2018-09-18
Fix code snippet in README

yanndupis:
Hello, 

Just a minor change in the code snippet. The code snippet wasn't running because we can't divide with a private tensor. 

Thanks,

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,11 @@ with tfe.protocol.Pond(server0, server1, crypto_producer) as prot:
     # get input from inputters as private values
     inputs = [prot.define_private_input(inputter) for inputter in inputters]
 
-    # securely sum all inputs and divide by count
-    result = reduce(lambda x,y: x+y, inputs) * (1 / len(inputs))
+    # get count inverse as a public value
+      cnt_inv = prot.define_public_variable(np.array([1 / len(inputs)]))
+
+    # sum all inputs and multiply by count inverse
+      result = reduce(lambda x, y: x + y, inputs) * cnt_inv
 
     # send result to receiver who can finally decrypt
     result_op = prot.define_output([result], result_receiver)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,7 +29,7 @@ with tfe.protocol.Pond(server0, server1, crypto_producer) as prot:
     inputs = [prot.define_private_input(inputter) for inputter in inputters]
 
     # securely sum all inputs and divide by count
-    result = reduce(lambda x,y: x+y, inputs) / len(inputs)
+    result = reduce(lambda x,y: x+y, inputs) * (1 / len(inputs))
```
mortendahl: want to implement (public) division as a fix instead? :D it shouldn't be that hard: you could simply take the inverse when the divisor is passed to `lift`

mortendahl: happy to help of course

README.md
```diff
@@ -76,6 +76,7 @@ Finally, we also loaded the pre-specified hostmap configuration from file using.
 # load host map configuration from file
 config = tfe.config.load('config.json')
 ```
+Take a look at [`/tools/gcp/link`](https://github.com/mortendahl/tf-encrypted/blob/master/tools/gcp/link) as an example to generate the config file for gcp. If you run it locally, you can use simply `tfe.LocalConfig`. You can find an example [here](https://github.com/mortendahl/tf-encrypted/blob/doc/examples/federated-average/run.py#L20).
```
mortendahl: fyi urls can be relative, ie `./tools/gcp/link` instead of `https://github.com/mortendahl/tf-encrypted/blob/master/tools/gcp/link` -- this would make the link work on non-master branches as well

yanndupis:
Hey @mortendahl, I have addressed public division, just not sure that's what you were expecting or if there is a better way to do it. Thanks

jvmancuso:
do we have tests for public division yet? could be good to add in, perhaps in a different PR?

yanndupis:
@jvmancuso sorry not 100% sure i follow you. The original code was `result = reduce(lambda x,y: x+y, inputs) / len(inputs)` and it was giving the following error `TypeError: unsupported operand type(s) for /: 'PondPrivateTensor' and 'int'`, do you want something else?

README.md
```diff
@@ -28,8 +28,11 @@ with tfe.protocol.Pond(server0, server1, crypto_producer) as prot:
     # get input from inputters as private values
     inputs = [prot.define_private_input(inputter) for inputter in inputters]
 
-    # securely sum all inputs and divide by count
-    result = reduce(lambda x,y: x+y, inputs) / len(inputs)
+    # get count inverse as a public value
+      cnt_inv = prot.define_public_variable(np.array([1 / len(inputs)]))
```
mortendahl: easy way: `reduce(lambda x, y: x + y, inputs) * (1 / len(inputs))`
best way: add support for public division to pond

mortendahl: jup, that works for me -- is that what's done in the actual code..?

yanndupis: In this PR, I currently have
```
# get count inverse as a public value
      cnt_inv = prot.define_public_variable(np.array([1 / len(inputs)]))

    # sum all inputs and multiply by count inverse
      result = reduce(lambda x, y: x + y, inputs) * cnt_inv
```

in the meantime I implement public division in pond, do you prefer this or `reduce(lambda x, y: x + y, inputs) * (1 / len(inputs))`

PR #115, opened 2018-09-18
Updates for Int32 tensor

mortendahl:


justin1121:
Great work! Just need to fix the build.

PR #116, opened 2018-09-18
Improve serving

morgangiraud:


tensorflow_encrypted/config.py
```diff
@@ -225,11 +245,13 @@ def session(
         master: Optional[Union[int, str]] = None,
         log_device_placement: bool = False
     ) -> tf.Session:
-
+        # cpu_cores = get_cpu_quota_within_docker() or multiprocessing.cpu_count()
```
ianlivingstone: If we're not using this function, we should probably remove it (and the commented out code).

morgangiraud: The issue linked in the comment is valuable for any future developer that would wonder if changing those values would lead to an improvement.

Useful pointers are better placed when they are right next to the code in my experience.

ianlivingstone: Is there a reason this is disable? Can we put it behind a flag or something similar instead of having it commented out?

morgangiraud: hum ... maybe the best thing would be to actually enable it and test it again. If it works then it is probably more robust with than without.

ianlivingstone: Deferred this for now into https://github.com/mortendahl/tf-encrypted/issues/124

PR #120, opened 2018-09-18
native binarize and some more tests around native tensors

justin1121:


tensorflow_encrypted/tensor/native.py
```diff
@@ -7,6 +7,8 @@
 from ..config import run
 
 INT_TYPE = tf.int32
+bits = 31
+p = 67
 
 
 class NativeTensor(object):
```
jvmancuso: Should NativeTensor be an abstract base class here? Unless there's a good reason for the opposite, it's usually good practice to do so.

justin1121: Do you have a good link explaining the benefits of ABC? I can't find anything concrete. I'm sure this can just be one though.

justin1121: As far as I can ABCs shouldn't have implementation.

justin1121: I'm going to merge this for now can figure it out later. I think this might come into play in the PR i'm working on right now.

PR #121, opened 2018-09-18
Gcp readme

yanndupis:
Just add two clarifications in the READMEs for gcp and mnist example with gcp.

Thanks,


yanndupis:
Should be good now, thanks,

Issue #92, opened 2018-09-06
Clean up Int100Tensor by moving to CrtTensor


mortendahl:

mortendahl:
Fixed previously

PR #128, opened 2018-09-18
introduce a factory pattern for creating tensors inside of a protocol

justin1121:
This PR creates an `AbstractFactory` and an `AbstractTensor` class so that we can take advantage of a factory type pattern to dynamically create tensors inside of any protocol. These classes are currently extended by both the Int100Tensor and then NativeTensor. I also had to move some functions like `_share`, `_encode`, etc. into the `Pond` protocol so that they could make use of this factory. Some globals and asserts were also moved into the `Pond` constructor.


tensorflow_encrypted/protocol/pond.py
```diff
@@ -52,25 +47,44 @@ def __init__(
...
         crypto_producer: Player,
-        use_noninteractive_truncation: bool=False
+        use_noninteractive_truncation: bool=False,
+        factory: AbstractFactory=Int100Factory()
     ) -> None:
         self.server_0 = server_0
         self.server_1 = server_1
         self.crypto_producer = crypto_producer
         self.use_noninteractive_truncation = use_noninteractive_truncation
+        self.factory = factory
```
mortendahl: could we call it something else, maybe simply `backing_tensor` or `tensor_factory`?

tensorflow_encrypted/protocol/pond.py
```diff
@@ -34,12 +32,9 @@
 TRUNCATION_GAP = 20
 BOUND = 2**30  # bound on magnitude of encoded numbers: x in [-BOUND, +BOUND)
 STATISTICAL_SECURITY = 40  # number of bit for statistical security TODO[Morten] write assertions
-
-# modulus
-M = BackingTensor.modulus
-
-assert log2(BITPRECISION_INTEGRAL + BITPRECISION_FRACTIONAL) <= log2(BOUND)
-assert log2(M) >= 2 * (BITPRECISION_INTEGRAL + BITPRECISION_FRACTIONAL) + log2(1024) + TRUNCATION_GAP
+K = None
```
mortendahl: let's move these three guys to Pond as instance fields instead!

tensorflow_encrypted/protocol/pond.py
```diff
@@ -966,8 +1016,8 @@ class PondPublicPlaceholder(PondPublicTensor):
     """
 
     def __init__(self, prot, placeholder_on_0, placeholder_on_1, is_scaled):
-        assert type(placeholder_on_0) is BackingPlaceholder, type(placeholder_on_0)
-        assert type(placeholder_on_1) is BackingPlaceholder, type(placeholder_on_1)
+        assert type(placeholder_on_0) is AbstractPlaceholder, type(placeholder_on_0)
```
mortendahl: probably `isinstance` here as well..?

tensorflow_encrypted/protocol/pond.py
```diff
@@ -986,9 +1036,9 @@ class PondPrivatePlaceholder(PondPrivateTensor):
     """
 
     def __init__(self, prot, placeholder, tensor0, tensor1, is_scaled):
-        assert type(placeholder) is BackingPlaceholder, type(placeholder)
-        assert type(tensor0) is BackingTensor, type(tensor0)
-        assert type(tensor1) is BackingTensor, type(tensor1)
+        assert type(placeholder) is AbstractPlaceholder, type(placeholder)
```
mortendahl: `isinstance`?

tensorflow_encrypted/tensor/factory.py
```diff
@@ -0,0 +1,29 @@
...
+    @abstractmethod
+    def Placeholder(self, shape: List[int]) -> AbstractPlaceholder:
+        pass
+
+    @abstractproperty
+    def modulus(self) -> int:
+        pass
+
+    @abstractmethod
+    def stack(self, xs: List[AbstractTensor], axis: int=0) -> AbstractTensor:
```
mortendahl: let's burn this guy at the stake in #92

tensorflow_encrypted/tensor/int100.py
```diff
@@ -59,7 +61,7 @@ class Int100Tensor(object):
     def __init__(
         self,
         native_value: Optional[Union[np.ndarray, tf.Tensor]],
-        decomposed_value: Optional[Union[List[np.ndarray], List[tf.Tensor]]] = None
+        decomposed_value: Optional[Union[List[np.ndarray], List[tf.Tensor], List[int]]] = None
```
mortendahl: why the `List[int]`?

justin1121: something was complaining not sure what anymore! will remove for now

mortendahl: i'm guessing strange things might happen if you pass a list of ints -- did you try?

tensorflow_encrypted/tensor/int100.py
```diff
@@ -80,10 +82,28 @@ def from_same(value: 'Int100Tensor') -> 'Int100Tensor':
         return Int100Tensor.from_decomposed(value.backing)
 
     @staticmethod
-    def from_decomposed(value: Union[List[np.ndarray], List[tf.Tensor]]) -> 'Int100Tensor':
+    def from_decomposed(value: Union[List[np.ndarray], List[tf.Tensor], List[int]]) -> 'Int100Tensor':
```
mortendahl: same here

tensorflow_encrypted/tensor/int100.py
justin1121: ints are getting passed in here: https://github.com/mortendahl/tf-encrypted/pull/128/files#diff-0697318ec50fe33b892f344d6d00b5a4L89 does the zeros function work?

mortendahl: ah good catch, that's probably a bug, should be wrapped in np.array i think

PR #130, opened 2018-09-19
quick readme fix

mortendahl:
Small update to the README to avoid variables etc

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,17 +30,13 @@ with tfe.protocol.Pond(server0, server1, crypto_producer) as prot:
     # get input from inputters as private values
     inputs = [prot.define_private_input(inputter) for inputter in inputters]
 
-    # get count inverse as a public value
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -81,7 +77,8 @@ Finally, we also loaded the pre-specified hostmap configuration from file using.
 # load host map configuration from file
 config = tfe.config.load('config.json')
 ```
-Take a look at [`/tools/gcp/link`](./tools/gcp/link) as an example to generate the config file for gcp. If you run it locally, you can use simply `tfe.LocalConfig`. You can find an example [here](./examples/federated-average/run.py#L20).
```
mortendahl: note that I removed line number reference here as things like that tend to change without being properly fixed up in these other places

mortendahl:
waiting for @yanndupis to take a look since he worked on this

yanndupis:
Look good to me!

PR #143, opened 2018-09-19
fixed links in the contents of the getting started doc

bendecoste:


bendecoste:
@yanndupis this was in the one you approved last but i moved it into a separate PR so that it could be its own commit.

Issue #136, opened 2018-09-19
Make sure we support TensorBoard histograms etc


mortendahl:
When e.g. mixing local training with private predictions it can be useful to know the distribution of values that the private predictions have to support, in order to e.g. understand where interval boundaries are broken. TensorBoard already supports gathering statistics like this so let's just make sure we support it. Might be as simple as running `summaries = tf.summary.merge_all()` in config.run

mortendahl:
see also https://www.tensorflow.org/guide/summaries_and_tensorboard

mortendahl:
leaning towards removing `config.run` and letting it be up to the user to take care of writing summaries etc.

for our own inspection purposes (ie what config does now) we can provide a small helper, potentially just renaming `config.run`

ianlivingstone:
Could we get an update as to why this is blocked?

yanndupis:
We wanted to test Tensorboard histograms on private tensors in the following program [netwok.a](https://github.com/mortendahl/tf-encrypted/blob/master/examples/securenn/network_a.py). However we discovered that there was some issues in the code, probably plaintext training #164. We will probably want to fix those before working on histogram.

mortendahl:
@yanndupis where do you think we are on this? Is it something we should bring focus back to?

mortendahl:
Great, moving into backlog

PR #148, opened 2018-09-20
Hack: SecureNN Experiment Benchmarks

jvmancuso:
Contains each of the models A, B, C, D from the SecureNN experiments at `examples/securenn/network_*.py`.  Most are based on the previous code from `examples/securenn/network_a.py`, updated to conform to best practices for weight initialization.  Waiting on #136 to confirm that the models train to expected accuracies in plaintext.

Current secure predictions are quite poor, suggesting either (1) a flaw in the original plaintext training code, (2) significant relu approximation errors, or (3) a bug in the pond operations.

My bet is on (2,) but maybe also (3) 😄

### DIFF
--- a/examples/securenn/network_a.py
+++ b/examples/securenn/network_a.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 import sys
+import math
 from typing import List
 
 import tensorflow as tf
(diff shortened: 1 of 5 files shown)


examples/securenn/network_a.py
```diff
@@ -25,14 +26,18 @@
 
 class ModelTrainer(tfe.io.InputProvider):
 
-    BATCH_SIZE = 30
+    BATCH_SIZE = 32
```
mortendahl: fyi 30 was picked to make sure all batches have the same size (ie divisor of 60k), not sure if that's a real problem

jvmancuso: nope! batch size affects the variance of the update, but not its expectation -- so having one batch be slightly more variant than the rest doesn't introduce any bias into the learning update

jvmancuso:
bumped it!

PR #151, opened 2018-09-20
Hack: logreg training example

### LINKED ISSUE #152: Multiply a PondPrivateTensor and a scalar from the left
Today we can do:
`out = xp * 1.`

But we can't do:
`out = 1. * xp`

morgangiraud:
close #152 

Note: Backprop is not working (the model is not training at all). I can't find any obvious error in my implementation so I think it's probably due to sigmoid approximation. 
More on this below:

The working TF code for the derivative looks like this:
```python
# Training model
out = tf.matmul(x, W) + b
pred = tf.sigmoid(out)

# Backprop
# dc_pred = - y / pred + (1 - y) / (1 - pred)
# dc_out = pred * (1 - pred) * dc_pred
# equivalent to:
dc_out = pred - y
```

So currently I apply the derivative equivalent in the encrypted setup:
```python
# Training model
out = prot.dot(xp, W) + b
pred = prot.sigmoid(out)

# Backprop
dc_out = pred - yp
```

Now, the thing is that this simplification is not true and probably too much off with our current approximation. 
Also even if I compute the real polynomial derivative to get `dc_out` from `dc_pred`, we need to divide by a private tensor (`pred`) to get `dc_pred` which is not yet possible.

What do you think? @jvmancuso @mortendahl

### DIFF
--- a/examples/logreg/data.py
+++ b/examples/logreg/data.py
@@ -1,0 +1,38 @@
+from typing import Tuple
+
+import numpy as np
+import tensorflow as tf
+
+
+def norm(x: tf.Tensor, y: tf.Tensor) -> Tuple[tf.Tensor, tf.Tensor]:
+    return tf.cast(x, tf.float32), tf.expand_dims(y, 0)
+
+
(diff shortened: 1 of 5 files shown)


mortendahl:
not sure what could be wrong

Issue #154, opened 2018-09-20
Create select share protocol


bendecoste:
child of #106

bendecoste:
Closed by #158

PR #158, opened 2018-09-20
Implement select share (secureNN) (#154)

bendecoste:
Implements select share in the secureNN protocol and adds a test

### DIFF
--- a/tensorflow_encrypted/protocol/pond.py
+++ b/tensorflow_encrypted/protocol/pond.py
@@ -914,6 +914,11 @@ class PondPrivateTensor(PondTensor):
     def reveal(self):
         return self.prot.reveal(self)
 
+    @staticmethod
+    def zero(prot: Protocol, shape: List[int]) -> 'PondPrivateTensor':
+        zero = np.zeros(shape=shape)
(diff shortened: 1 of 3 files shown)


tensorflow_encrypted/protocol/pond.py
```diff
@@ -914,6 +914,11 @@ def unwrapped(self):
     def reveal(self):
         return self.prot.reveal(self)
 
+    @staticmethod
+    def zero(prot: Protocol, shape: List[int]) -> 'PondPrivateTensor':
```
bendecoste: Unsure about doing this this way.  I wanted a zero method as a PondVariable though as I don't think we should be forced to use Int100Tensors with PondTensors.

justin1121: I think related to the work I'm doing now there will be a static method on the `AbstractTensor` that will force subclasses of `AbstractTensor` to implement the zeros function for the given tensor.

bendecoste:
> Looks great, maybe add an issue about updating the zeros call once we have the Abstract tensor etc. in place.

Created #159.

PR #166, opened 2018-09-21
int 32 binarize

justin1121:
Implements binarize for int32. Numbers of bits to generate and p are hardcoded right now. Might want to change that and binarize can take a p arg.

### DIFF
--- a/tensorflow_encrypted/tensor/int32.py
+++ b/tensorflow_encrypted/tensor/int32.py
@@ -116,13 +116,12 @@ class Int32Tensor(AbstractTensor):
         return Int32Tensor.from_native(tf.concat(backing, axis=axis))
 
     def binarize(self) -> NativeTensor:
-        """Computes bit decomposition of tensor
-         tensor: ndarray of shape (x0, ..., xn)
-        returns: a binary tensor of shape (x0, ..., xn, bits) equivalent to tensor
-        """
         bitwidths = tf.range(bits, dtype=tf.int32)
-        for i in range(len(self.shape)):
-            bitwidths = tf.expand_dims(bitwidths, 0)
+
+        final_shape = [1] * len(self.shape)
+        final_shape.append(bits)
+
+        bitwidths = tf.reshape(bitwidths, final_shape)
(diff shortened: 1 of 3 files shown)


tests/test_int32_tensor.py
```diff
@@ -27,6 +27,19 @@ def test_pond(self) -> None:
                 out = z.reveal().eval(sess)
                 np.testing.assert_array_almost_equal(out, [4, 4], decimal=3)
 
+    def test_binarize(self) -> None:
+        x = Int32Tensor(tf.constant([2**31 + 2**31 + 3, 3], dtype=np.int32))
```
jvmancuso: can we add one more test with randomly generated values?  should be robust, feel like this test is useful but pretty specific

tensorflow_encrypted/tensor/int32.py
```diff
@@ -111,6 +115,19 @@ def concat(x: List['Int32Tensor'], axis: int) -> 'Int32Tensor':
 
         return Int32Tensor.from_native(tf.concat(backing, axis=axis))
 
+    def binarize(self) -> NativeTensor:
+        """Computes bit decomposition of tensor
```
mortendahl: this doesn't seem up to date, eg taking ndarray as input, remove for now?

tensorflow_encrypted/tensor/int32.py
```diff
@@ -111,6 +115,19 @@ def concat(x: List['Int32Tensor'], axis: int) -> 'Int32Tensor':
...
+         tensor: ndarray of shape (x0, ..., xn)
+        returns: a binary tensor of shape (x0, ..., xn, bits) equivalent to tensor
+        """
+        bitwidths = tf.range(bits, dtype=tf.int32)
+        for i in range(len(self.shape)):
```
mortendahl: think this could be replaced by `tf.reshape(bitwidths, [1]*len(..))`, don't know if it's faster though

tensorflow_encrypted/tensor/int32.py
justin1121: good catch

PR #168, opened 2018-09-24
Add docker thread checking

### LINKED ISSUE #124: Re-visit explicitly setting the number of threads during configuration
Reference https://github.com/mortendahl/tf-encrypted/pull/116/files#r218417607

Currently, the `cpu_cores` code is uncommented. We should revisit this and see if its necessary and then decide whether it should be a configurable option, automated, or removed entirely.

morgangiraud:
Closes #124 

I also added the capacity to download a model from a public URL. Useful feature to avoid hardcoding the model inside a docker container.

### DIFF
--- a/bin/serve
+++ b/bin/serve
@@ -4,10 +4,13 @@ import sys
 import os
 import uuid
 import time
+import re
+import argparse
+from pathlib import Path
+from urllib import request as urlDownloader
 from threading import Thread
 from typing import Dict, Any, Tuple, Callable, List
 
-import argparse
 import tensorflow as tf
 from tensorflow.python.platform import gfile
 import numpy as np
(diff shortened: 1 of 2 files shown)


tensorflow_encrypted/config.py
```diff
@@ -245,13 +247,15 @@ def session(
         master: Optional[Union[int, str]] = None,
         log_device_placement: bool = False
     ) -> tf.Session:
-        # cpu_cores = get_cpu_quota_within_docker() or multiprocessing.cpu_count()
+        cpu_cores = get_cpu_quota()
```
mortendahl: should we only specify this if we're inside docker or is it generally useful (got the impression that TF is already making some choices here)?

bin/serve
```diff
@@ -73,21 +76,38 @@ else:
...
+        "--model_url", type=str, default='', help="Model's public URL")
     config = parser.parse_args()
 
-    print('Loading model: %s' % config.model_name)
     model_file = os.path.join(
         os.path.dirname(os.path.realpath(__file__)),
         "..",
         "models",
-        "%s.pb" % config.model_name,
+        "{}.pb".format(config.model_name),
```
mortendahl: 🎉

bin/serve
```diff
@@ -73,21 +76,38 @@ else:
...
+        regex = re.compile(
+            r'^(?:http|ftp)s?://'  # http:// or https://
+            r'(?:(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.)+(?:[A-Z]{2,6}\.?|[A-Z0-9-]{2,}\.?)|'  # domain...
+            r'localhost|'  # localhost...
+            r'\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})'  # ...or ip
+            r'(?::\d+)?'  # optional port
+            r'(?:/?|[/?]\S+)$', re.IGNORECASE
+        )
+        if re.match(regex, config.model_url) is not None:
+            print('Downloading model')
```
mortendahl: nice

tensorflow_encrypted/config.py
morgangiraud: Now that you are mentioning it, we should probably start by avoiding setting it outside docker until we have a strong understanding of possible related optimizations.

I'll update this part so it returns `None` outside of docker.

tensorflow_encrypted/config.py
```diff
@@ -247,9 +246,9 @@ def session(
         master: Optional[Union[int, str]] = None,
         log_device_placement: bool = False
     ) -> tf.Session:
-        cpu_cores = get_cpu_quota()
+        cpu_cores = get_docker_cpu_quota()
```
mortendahl: is this enough to only specify the properties if running in docker? seems like the protobuf file gives default integer values..?

morgangiraud: Not sure to follow, concerning TF config, if `None` (or no values) are provided, TF will optimize them itself.

mortendahl: could we avoid setting them instead of passing None? or are we sure that's doing what we expect?

morgangiraud: Interesting fact: the default value is 0 since it's should be a `int`.

I'll this piece of code so we don't set it with `None` then

PR #173, opened 2018-09-24
Python35

### LINKED ISSUE #167: PyPi support for Python 3.5
Python 3.5 is what currently comes with Debian 9 (which is popular on GCP and Raspberry Pi), and it's messy to install 3.6. At the same time the only reason we don't currently support 3.5 seems to be the type hints, so maybe something as simple as [`strip-hints`](https://pypi.org/project/strip-hints/) could be used to build a 3.5 compatible PyPi package?

cc @ianlivingstone @morgangiraud any thoughts?

mortendahl:
closes #167 

- removed some type hints, yet all of them can easily be inferred directly so don't think this makes a big difference

- fixed a few indentation bugs in Pond

- memoized a few methods in Pond

- lowered the required TF version to 1.9 to get it running on RPIs

note: I tried to get `strip-hints` to work, but it appears to be a runtime thing only that didn't make me convinced

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ cd tf-encrypted
 pip install -e .
 ```
 
-Note however that currently **only Python 3.6 is supported**; to manage this we recommend using a package manager like pip or conda.
+Note however that currently **only Python 3.5 and 3.6 are supported**; to manage this we recommend using a package manager like pip or conda.
 
 After successful installation you should be able to e.g. run the examples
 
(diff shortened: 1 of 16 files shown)


ianlivingstone:
Do we have anything (e.g. CI) that ensures what we produce continues to work in 3.5 vs 3.6?

mortendahl:
No and I think it's already broken actually

bendecoste:
That's like a one liner on travis .. assuming it's similar with circle ci!

PR #177, opened 2018-09-25
 implement a shared conv2d for int32 and int64

justin1121:
Implement a shared conv2d for int32 and int64. It has also been made generic enough to work with int100/crt tensor. This change will be coming down the line next.

### DIFF
--- a/tensorflow_encrypted/tensor/int100.py
+++ b/tensorflow_encrypted/tensor/int100.py
@@ -115,7 +115,7 @@ class Int100Tensor(AbstractTensor):
         evaluated_backing = run(sess, self.backing, feed_dict=feed_dict, tag=tag)
         return Int100Tensor.from_decomposed(evaluated_backing)
 
-    def to_int32(self) -> Union[tf.Tensor, np.ndarray]:
+    def to_native(self) -> Union[tf.Tensor, np.ndarray]:
         return _crt_recombine_explicit(self.backing, 2**31)
 
     def to_bigint(self) -> np.ndarray:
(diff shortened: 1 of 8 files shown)


tensorflow_encrypted/tensor/int100.py
```diff
@@ -115,7 +115,7 @@ def eval(self, sess: tf.Session, feed_dict: Dict[Any, Any]={}, tag: Optional[str
         evaluated_backing = run(sess, self.backing, feed_dict=feed_dict, tag=tag)
         return Int100Tensor.from_decomposed(evaluated_backing)
 
-    def to_int32(self) -> Union[tf.Tensor, np.ndarray]:
+    def to_native(self) -> Union[tf.Tensor, np.ndarray]:
```
mortendahl: not sure what's best here, but you might be right that it should be `native`. my issue is that the code implicitly chops off anything outside the int32 range, and is converting numbers into that dtype. 

what was the need for renaming it?

tensorflow_encrypted/tensor/int64.py
```diff
@@ -45,39 +46,40 @@ def __repr__(self) -> str:
...
         return Int64Tensor(x.value * y.value)
 
-    def dot(self, other) -> 'Int64Tensor':
+    def dot(self, other: Any) -> 'Int64Tensor':
         x, y = Int64Tensor.lift(self), Int64Tensor.lift(other)
-        return Int64Tensor(tf.matmul(x.value, y.value))
 
-    def im2col(self, h_filter, w_filter, padding, strides) -> 'Int64Tensor':
-        raise NotImplementedError()
+        print(x.value.dtype)
```
mortendahl: leftovers?

tensorflow_encrypted/tensor/int100.py
justin1121: All the tensors need to share a similar function since the `to_int32` function is being used when decoding the numbers in pond. Didn't want it to be called `to_int32` in the `Int64Tensor`.

mortendahl: yeah make sense. there is something underlying here we haven't solved yet but it can be pushed till later when we refactor/figure out what to do with the different tensors

Issue #149, opened 2018-09-20
Simplified interface closer to typical TF scripts


mortendahl:
We could expose an interface closer to how TF is used in many scripts.

For instance, the following may not be that hard to get working

```python
@tfe.public_input('model-owner')
def get_model():
    return tf.constant([1,2,3,4])

@tfe.private_input('prediction-client')
def get_input():
    return tf.constant([4,3,2,1])

@tfe.output('prediction-client')
def reveal_output(y):
    return tf.Print([], [y])

w = get_model()

x = get_input()

# secure computation
y = tfe.matmul(w, x)

pred_op = reveal_output(y)

with config.session() as sess:
    sess.run(pred_op)
```

and, if Python allows, could also be slightly rewritten as

```python
def get_model():
    return tf.constant([1,2,3,4])

def get_input():
    return tf.constant([4,3,2,1])

def reveal_output(y):
    return tf.Print([], [y])

@tfe.public_input('model-owner')
w = get_model()

@tfe.private_input('prediction-client')
x = get_input()

# secure computation
y = tfe.matmul(w, x)

@tfe.output('prediction-client')
pred_op = reveal_output(y)

with config.session() as sess:
    sess.run(pred_op)
```

mortendahl:
some issues:
- does this help train the user to think in terms of several players
- how to make creation of protocol instance implicit (with a default)
- how to avoid passing config instances around everywhere

ianlivingstone:
@morgangiraud can we close this one up with #175?

morgangiraud:
Soon!

mortendahl:
Closing, seems to have been fixed along the way

PR #182, opened 2018-09-26
Improve tf-API

### LINKED ISSUE #175: Make the API closer to TF
We want to have an improved API for TFE. The goal is to be as close as possible to TF, so people don't have to learn another API on top of TF.

More precisely this issue focus on transforming the [TFE logistic regression example](https://github.com/mortendahl/tf-encrypted/blob/master/examples/logreg/tfe-logreg.py) so it looks the same as the [TF example](https://github.com/mortendahl/tf-encrypted/blob/master/examples/logreg/tf-logreg.py).

This include:
- [x] Add `reduce_sum` alias
- [x] Add `matmul` alias
- [x] Add a default LocalConfig that will be used by a protocol (Add a warning)
- [x] Add a new sessions system
- [x] Add a default protocol that will be used by default when calling operations (tfe.matmul ...)
- [x] Allows `define_private_input`to be used as an annotation

morgangiraud:
connect #175 

@mortendahl I want to move step by step here. I just finished allowing `define_private_input` to handle a `tf.Tensor` input. What do you think?

### DIFF
--- a/tensorflow_encrypted/protocol/pond.py
+++ b/tensorflow_encrypted/protocol/pond.py
@@ -337,7 +337,7 @@ class Pond(Protocol):
             return share0 + share1
 
     @memoize
-    def assign(self, variable, value):
+    def assign(self, variable: 'PondPrivateVariable', value) -> tf.Operation:
         assert isinstance(variable, PondPrivateVariable), type(variable)
         assert isinstance(value, PondPrivateTensor), type(value)
         assert variable.is_scaled == value.is_scaled, "Scaling must match: {}, {}".format(variable.is_scaled, value.is_scaled)
@@ -408,7 +408,7 @@ class Pond(Protocol):
             raise TypeError("Don't know how to lift {}, {}".format(type(x), type(y)))
 
     @memoize
-    def sum(self, x, axis, keepdims):
+    def sum(self, x, axis=None, keepdims=None):
         x = self.lift(x)
 
         dispatch = {
@@ -421,6 +421,9 @@ class Pond(Protocol):
             raise TypeError("Don't know how to sum {}".format(type(x)))
 
         return func(self, x, axis, keepdims)
+
+    def reduce_sum(self, x, axis=None, keepdims=None):
+        return self.sum(x, axis, keepdims)
 
     @memoize
     def sub(self, x, y):
@@ -815,8 +818,11 @@ class PondTensor(abc.ABC):
     def __add__(self, other):
         return self.prot.add(self, other)
 
-    def sum(self, axis, keepdims=False):
+    def sum(self, axis=None, keepdims=None):
         return self.prot.sum(self, axis, keepdims)
+
+    def reduce_sum(self, axis=None, keepdims=None):
+        return self.sum(self, axis, keepdims)
 
     def sub(self, other):
         return self.prot.sub(self, other)
@@ -1517,8 +1523,8 @@ def _add_masked_masked(prot, x, y):
 
 def _sum_core(prot: Pond,
               x: PondTensor,
-              axis: int,
-              keepdims: Optional[bool]) -> Tuple[AbstractTensor, AbstractTensor]:
+              axis: Optional[int] = None,
+              keepdims: Optional[bool] = None) -> Tuple[AbstractTensor, AbstractTensor]:
 
     x_on_0, x_on_1 = x.unwrapped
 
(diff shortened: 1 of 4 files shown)


examples/logreg/tfe-logreg.py
```diff
@@ -43,27 +34,31 @@ def consume(v: Union[np.ndarray, tf.Tensor], prot: tfe.protocol.Pond) -> PondPri
 
     W = prot.define_private_variable(tf.random_uniform([nb_feats, 1], -0.01, 0.01))
     b = prot.define_private_variable(tf.zeros([1]))
-    xp = consume(x, prot)  # We secure our inputs
-    yp = consume(y, prot)
+    xp = prot.define_private_input(x)  # We secure our inputs
```
mortendahl: not convinced by this tbh, we're opening up for security issues (even if a warning is printed) and in this case it doesn't seem needed since both `x` and `y` are the output of a function (see my suggestion in https://github.com/mortendahl/tf-encrypted/issues/149)

examples/logreg/tfe-logreg.py
```diff
@@ -43,27 +34,31 @@ def consume(v: Union[np.ndarray, tf.Tensor], prot: tfe.protocol.Pond) -> PondPri
...
+    yp = prot.define_private_input(y)
 
     # Training model
-    out = prot.dot(xp, W) + b
+    out = prot.matmul(xp, W) + b
```
mortendahl: 👍

examples/logreg/tfe-logreg.py
```diff
@@ -43,27 +34,31 @@ def consume(v: Union[np.ndarray, tf.Tensor], prot: tfe.protocol.Pond) -> PondPri
...
+    db = prot.reduce_sum(1. * dc_out, axis=0)
     ops = [
         prot.assign(W, W - dW * learning_rate),
         prot.assign(b, b - db * learning_rate)
     ]
 
     # Testing model
-    xp_test = consume(x_test, prot)  # We secure our inputs
-    yp_test = consume(y_test, prot)
+    xp_test = prot.define_private_input(x_test)  # We secure our inputs
```
mortendahl: same here, `x_test` and `y_test` are return values from function call

examples/logreg/tfe-logreg.py
morgangiraud: I’ve been thinking about the current problem for `define_private_input`, I came to the conclusion that the name itself is misleading. `define_private_input` is not really defining anything, it secures a provider for now. `prot.secure_provider` would be a better name 

I think one of the most complicated thing to understand (from a new user point of view) is that you want "the secure step" to happens on the machine that provide the input but you can define the code that provide the input on the machine running the main script.
We achieve that by creating a function that load the inputs and use TF to execute the function on a different  machine, in the meantine we sneak in out secure step by wrapping the function outputs. All of that led to the notion of InputProvider probably.

I would argue that it would be better to have directly:
```python
def secure_input_function(func, player=default_input_player):
...
```
I don't like to have objects for everything, the more we can have "pure function" the better we are.

Side note: I would agree with you that this can be a decoration.

What do you think?

mortendahl: renaming to something like `private_input` would make sense to me, but `secure_provider` less so ("private input"/"public input" is somewhat normal in MPC literature).

letting the input method on `prot` take either objects or functions with players could both work, but I think decorators is actually better as it not only allows us to hide this stuff, it also gathers the function definition and where it's executed, as well as make it easy to use default prot objects:
```python
@tfe.private_input('prediction-client')
def get_input():
    return tf.constant([4,3,2,1])

x = get_input()
```
instead of
```python
def get_input():
    return tf.constant([4,3,2,1])

with ... as prot:
   x = prot.private_input(get_input, "prediction-client")
```

morgangiraud: Interesting.
From my engineer point of view I would rather see:
```python
@tfe.run_on('prediction-client')
def get_input():
    return tf.constant([4,3,2,1])
```
or maybe
```python
@tfe.secure_on('prediction-client')
def get_input():
    return tf.constant([4,3,2,1])
```

mortendahl: yeah I see what you mean. you need to specify if you want it as private or public values though (`PondPrivateTensor`/`PondPublicTensor`), at least I think it would be useful to be able to have both

mortendahl: any thought @justin1121 @bendecoste @jvmancuso @ianlivingstone ?

justin1121: I think we need to be able to define both public/private inputs this way. So I think I like `@tfe.private_input('prediction-client')` the best. 

Although, we could also pass a public/private flag into the `secure_on` function. Like `@tfe.secure_on('prediction-client', private=True)`. But, then the `secure_on` name doesn't make sense if it's public. Maybe `provide` would be better.

Think I'm on the side of `private_input`, in the end. Its an interesting balance between what would make sense for the average user and what would make sense to someone who knows a lot about MPC.

PR #183, opened 2018-09-26
181 sum bug

### LINKED ISSUE #181: reduce_sum bug

morgangiraud:
Closes #181 

@jvmancuso Is it you who wrote the code of `crt_sum`? 

I found a bug on it (you can check the test file I added). It is due to the following code:
```python
def crt_sum(x, axis, keepdims=None):
        with tf.name_scope('crt_sum'):
            dims = x[0].shape.dims.copy()  #<--copy dims
            ax_len = dims.pop(axis) # Removing last dimensions from dims
            begins = [0] * len(dims) # countings dims after removing an element, for a 1d tensor this leads to begins = []
            ends = [x.value for x in dims]
            perm = [axis, *[i for i in range(len(dims) + 1) if i != axis]]
            x = [tf.transpose(xi, perm=perm) for xi in x]
            y = [tf.zeros(dims, dtype=xi.dtype) for xi in x]
            for i in range(ax_len):
                for j in range(len(x)):
                    sl = x[j][i]
                    y[j] += sl
                    y[j] %= m[j]
                begins[0] += 1 # does not exist for a 1d tensor
                ends[0] += 1
            if keepdims:
                return [tf.expand_dims(yi, axis) for yi in y]
            return y
```

My question is: why are you poping that element? the code looks very complicated for a sum, is there any specific reason?

### DIFF
--- a/tensorflow_encrypted/protocol/pond.py
+++ b/tensorflow_encrypted/protocol/pond.py
@@ -446,7 +446,7 @@ class Pond(Protocol):
         return z
 
     def reduce_sum(self, x, axis=0, keepdims=False):
-        return self.sum(x, axis=0, keepdims=False)
+        return self.sum(x, axis=axis, keepdims=False)
 
     @memoize
     def sub(self, x, y):
--- a/tensorflow_encrypted/tensor/crt.py
+++ b/tensorflow_encrypted/tensor/crt.py
@@ -260,20 +260,7 @@ def gen_crt_mod(m, int_type):
 def gen_crt_sum(m):
     def crt_sum(x, axis, keepdims=None):
         with tf.name_scope('crt_sum'):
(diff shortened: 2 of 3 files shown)


jvmancuso:
The bug you mentioned can be fixed by switching from `pop` to a simple `__getitem__`, but the larger issue is that this code looks like it was merged prematurely, since the `begins` and `ends` variables aren't used.    It was only meant to be as complex as needed to be able to handle summing over a dimension and to be able to `keepdims`.  Looks like that should be achievable using `tf.reduce_sum` in combination with `tf.expand_dims` (don't think I realized `reduce_sum` was there originally).  I'm testing that fix now 🙂

jvmancuso:
Deferred to #185.

PR #185, opened 2018-09-26
Extract reduce_sum contribution

### LINKED ISSUE #181: reduce_sum bug

morgangiraud:
Closes #181

### DIFF
--- a/tensorflow_encrypted/protocol/pond.py
+++ b/tensorflow_encrypted/protocol/pond.py
@@ -336,7 +336,7 @@ class Pond(Protocol):
             return share0 + share1
 
     @memoize
-    def assign(self, variable, value):
+    def assign(self, variable: 'PondPrivateVariable', value) -> tf.Operation:
         assert isinstance(variable, PondPrivateVariable), type(variable)
         assert isinstance(value, PondPrivateTensor), type(value)
         assert variable.is_scaled == value.is_scaled, "Scaling must match: {}, {}".format(variable.is_scaled, value.is_scaled)
@@ -407,7 +407,7 @@ class Pond(Protocol):
             raise TypeError("Don't know how to lift {}, {}".format(type(x), type(y)))
 
     @memoize
-    def sum(self, x, axis, keepdims):
+    def sum(self, x, axis=0, keepdims=False):
         x = self.lift(x)
 
         dispatch = {
@@ -420,6 +420,9 @@ class Pond(Protocol):
             raise TypeError("Don't know how to sum {}".format(type(x)))
 
         return func(self, x, axis, keepdims)
+
+    def reduce_sum(self, x, axis=0, keepdims=False):
+        return self.sum(x, axis=0, keepdims=False)
 
     @memoize
     def sub(self, x, y):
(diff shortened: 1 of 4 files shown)


tensorflow_encrypted/protocol/pond.py
```diff
@@ -421,6 +421,9 @@ def sum(self, x, axis, keepdims):
 
         return func(self, x, axis, keepdims)
 
+    def reduce_sum(self, x, axis=0, keepdims=False):
```
mortendahl: should we make default values matching TF or will that complicate matters? (eg https://www.tensorflow.org/api_docs/python/tf/reduce_sum)

tensorflow_encrypted/protocol/pond.py
```diff
@@ -818,9 +821,12 @@ def add(self, other):
     def __add__(self, other):
         return self.prot.add(self, other)
 
-    def sum(self, axis, keepdims=False):
+    def sum(self, axis=0, keepdims=False):
```
mortendahl: if TF only has `reduce_sum` then maybe we should simply follow that to keep it simple?

morgangiraud: depends if you want to be friendly with numpy users?

tensorflow_encrypted/protocol/pond.py
morgangiraud: I couldn't due on the complicated implementation of `crt_sum` under the hood. But if we replace everything with `tf.reduce_sum` under the hood then yes.
@jvmancuso ?

tensorflow_encrypted/protocol/pond.py
mortendahl: wdyt?

mortendahl: could we swap so that `reduce_sum` is called by `sum` at least? (okay this is vanity, I know..)

morgangiraud: I don't mind having aliases, but it needs to be mentioned in the doc so people don't think it's a different behavior under the hood.

mortendahl: that was my concern as well. my vote is on sticking with TF (since we're promoting that anyway) but okay with catering a bit for numpy users as well if you guys think it's worth it

tensorflow_encrypted/protocol/pond.py
```diff
@@ -1538,24 +1544,24 @@ def _sum_core(prot: Pond,
...
 
 def _sum_masked(prot: Pond,
                 x: PondMaskedTensor,
-                axis: int,
-                keepdims: Optional[bool]) -> PondPrivateTensor:
+                axis: Optional[int] = None,
+                keepdims: Optional[bool] = None) -> PondPrivateTensor:
     # y_on_0, y_on_1 = _sum_core(prot, x.unmasked, axis, keepdims)
     # return PondPrivateTensor(prot, y_on_0, y_on_1)
     raise NotImplementedError
```
mortendahl: for masked values, just forward to the unmasked values: `prot.sum(x.unmasked, .., ..)`

tests/test_reduce_sum.py
```diff
@@ -0,0 +1,58 @@
+import unittest
+
+import numpy as np
+import tensorflow as tf
+import tensorflow_encrypted as tfe
+
+
+class Testconcat(unittest.TestCase):
```
jvmancuso: From #183, we should rename this

morgangiraud: Good catch, I hate when people don't change it ... Now I'm hating myself 🥇

PR #187, opened 2018-09-26
Remove rsqrt

justin1121:


mortendahl:
FYI we can have square roots on public values, we just need to convert them away from CRT first, ie call something like `to_native()`, followed by a cast to floats, and then back to PondPublicTensor and CRT

PR #195, opened 2018-09-27
Add a default configuration

morgangiraud:
Connect #175 

A proposal for a default configuration

### DIFF
--- a/tensorflow_encrypted/protocol/securenn.py
+++ b/tensorflow_encrypted/protocol/securenn.py
@@ -1,23 +1,26 @@
+from __future__ import absolute_import
+from typing import Optional
 from .protocol import memoize
 from ..protocol.pond import (
     Pond, PondTensor
 )
 from ..player import Player
+from ..config import get_default_config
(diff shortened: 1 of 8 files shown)


tensorflow_encrypted/protocol/securenn.py
```diff
@@ -1,23 +1,26 @@
...
+        server_0: Optional[Player] = None,
+        server_1: Optional[Player] = None,
+        server_2: Optional[Player] = None,
         **kwargs
     ) -> None:
         super(SecureNN, self).__init__(
-            server_0=server_0,
-            server_1=server_1,
-            crypto_producer=server_2,
+            server_0=server_0 or get_default_config().get_player('server0'),
```
mortendahl: note you don't need the `or get_default...` part here, can just pass on None and let Pond handle it

morgangiraud: If you don't mind, I'll refactor that later 😊

PR #196, opened 2018-09-27
Add tfe session

morgangiraud:


tensorflow_encrypted/config.py
```diff
@@ -56,6 +57,103 @@ def get_docker_cpu_quota() -> Optional[int]:
     return cpu_cores
 
 
+class TFESession():
```
justin1121: Can we subclass `tf.Session`?

tensorflow_encrypted/config.py
```diff
@@ -56,6 +57,103 @@ def get_docker_cpu_quota() -> Optional[int]:
...
+        self.sess = sess
+
+    def __enter__(self):
+        self.sess = self.sess.__enter__()
+        return self
+
+    def __exit__(self, exec_type, exec_value, exec_tb):
+        self.sess.__exit__(exec_type, exec_value, exec_tb)
+
+    def sanitize_fetches(self, fetches: Any) -> List[Union[tf.Tensor, tf.Operation]]:
```
justin1121: What are `sanitize_fetches` and `decode_fetches` functions doing exactly? They're exposing hidden details of the underlying protocols and tensor types. For example, not all tensors currently have `.backing` members or `.from_decomposed` functions. Maybe they should but I think we should be leaving the decoding up to the protocol.

tensorflow_encrypted/config.py
morgangiraud: It's more of a wrapper around a `tf.session` currently. 

Do you think we can gain something from inheritance?

justin1121: Not a whole lot. Just thought we wouldn't have to implement the `__enter__` and `__exit__` methods. Not a big deal either way.

justin1121: Maybe that wouldn't even work. NVM!!

morgangiraud: maybe, I will try

tensorflow_encrypted/config.py
morgangiraud: Yes, indeed.
I wanted some feedback on how to generalize the behavior across all protocols

justin1121: I think I see what you're doing now. Not sure about the sanitized function but I think for decode we just need to expose a function in the protocol that knows how to decode. Sort of like the `.eval()` function on the `PondPublicTensor`.

tensorflow_encrypted/__init__.py
```diff
@@ -9,8 +11,11 @@
 from . import convert
 
 
+def set_random_seed(seed: int) -> None:
```
mortendahl: maybe we just call `tf.set_random_seed` directly if there's nothing more to be done?

tensorflow_encrypted/session.py
```diff
@@ -0,0 +1,135 @@
...
+
+__TFE_STATS__ = bool(os.getenv('TFE_STATS', False))
+__TFE_TRACE__ = bool(os.getenv('TFE_TRACE', False))
+__TFE_DEBUG__ = bool(os.getenv('TFE_DEBUG', False))
+__TENSORBOARD_DIR__ = str(os.getenv('TFE_STATS_DIR', '/tmp/tensorboard'))
+
+_run_counter = defaultdict(int)  # type: Any
+
+
+class TFESession():
```
mortendahl: rename to `Session` and get rid of `Session` method?

tests/test_share.py
```diff
@@ -23,23 +19,23 @@ def setUp(self):
 
     def test_share(self):
 
-        with tfe.protocol.Pond(*self.config.get_players('server0, server1, crypto_producer')) as prot:
+        with tfe.protocol.Pond(tfe.get_default_config().get_players('server0, server1, crypto_producer')) as prot:
```
mortendahl: could you just call `tfe.protocol.Pond()` (and all similar places)?

tensorflow_encrypted/session.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+    Wrap a Tensorflow Session
+    """
+
+    def __init__(self, sess: tf.Session) -> None:
```
mortendahl: what is `TFESession` derived from `tf.Session`, would it work? then you wouldn't have to forward these three methods

tensorflow_encrypted/session.py
```diff
@@ -0,0 +1,135 @@
...
+        self.sess = sess
+
+    def __enter__(self) -> 'TFESession':
+        self.sess = self.sess.__enter__()
+        return self
+
+    def __exit__(self, exec_type, exec_value, exec_tb):
+        self.sess.__exit__(exec_type, exec_value, exec_tb)
+
+    def sanitize_fetches(self, fetches: Any) -> Union[List[Any], tf.Tensor, tf.Operation]:
```
mortendahl: maybe rewrite this as a recursive function, think that would clean it up a lot and allow for deeper nesting

tensorflow_encrypted/session.py
```diff
@@ -0,0 +1,135 @@
...
+
+            writer.add_run_metadata(run_metadata, session_tag)
+            writer.close()
+
+            if __TFE_TRACE__ or write_trace:
+                chrome_trace = timeline.Timeline(run_metadata.step_stats).generate_chrome_trace_format()
+                with open('{}/{}.ctr'.format(__TENSORBOARD_DIR__, session_tag), 'w') as f:
+                    f.write(chrome_trace)
+
+        return self.decode_fetches(fetches, fetches_out)
```
mortendahl: you should be able to decode before fecthing

tensorflow_encrypted/session.py
```diff
@@ -0,0 +1,135 @@
...
+                        sanitized_fetches[idx].append(sub_fetch)
+            else:
+                if not isinstance(fetch, tf.Tensor) and not isinstance(fetch, tf.Operation):
+                    sanitized_fetches.append(fetch.value_on_0.backing)
+                else:
+                    sanitized_fetches.append(fetch)
+
+        return sanitized_fetches
+
+    def decode_fetches(self, fetches: Any, fetches_out: Any) -> List[Union[tf.Tensor, tf.Operation]]:
```
mortendahl: same here, looks like a recursive function! note remark later that you could probably decode during sanitizing

tensorflow_encrypted/__init__.py
morgangiraud: It depends how far we want to play along the lines of "You don't even need to import Tensorflow".

After this session PR, my goal is to find a way to duplicate all function protocols to the main scope (ie `prot.matmul` -> `tfe.matmul`).
The same work could be done to move all `tf` function to `tfe` scope, do you find that interesting or useless?

mortendahl: hmm, not sure. i like the idea of having the same API but maybe holding back a bit since:
- 1) not sure it's priority right now
- 2) we are complaining good integration with ordinary TF, ie in some use cases we might actually want people to import both

mortendahl: in any case, i like the idea of `tfe.matmul` for ops we have already a lot

tensorflow_encrypted/session.py
morgangiraud: I've reworked the dependencies between config and sessions. Feel cleaner now 👍

tensorflow_encrypted/session.py
morgangiraud: Done

tensorflow_encrypted/session.py
morgangiraud: Done

tensorflow_encrypted/session.py
morgangiraud: Not sure why, but if I try to do that, some examples start to hang.

tensorflow_encrypted/protocol/pond.py
```diff
@@ -895,9 +897,8 @@ def shape(self) -> List[int]:
     def unwrapped(self) -> Tuple[AbstractTensor, ...]:
         return (self.value_on_0, self.value_on_1)
 
-    def eval(self, sess, feed_dict={}, tag=None) -> np.ndarray:
-        value = self.value_on_0.eval(sess, feed_dict=feed_dict, tag=tag)
-        return self.prot._decode(value, self.is_scaled)
+    def decode(self) -> Union[np.ndarray, tf.Tensor]:
```
mortendahl: fyi this is a small anti-pattern as we normally have all functionality on the protocol class, and mostly use the tensor objects as structs

tensorflow_encrypted/session.py
```diff
@@ -0,0 +1,116 @@
...
+
+        if isinstance(fetches, (list, tuple)):
+            return [self.sanitize_fetches(fetch) for fetch in fetches]
+        else:
+            if isinstance(fetches, (tf.Tensor, tf.Operation)):
+                return fetches
+            elif isinstance(fetches, PondPublicTensor):
+                return fetches.decode()
+            elif isinstance(fetches, AbstractTensor):
+                return fetches.to_native()
```
mortendahl: 👍

tensorflow_encrypted/protocol/pond.py
morgangiraud: I see but I can't help to find very strange to write this `t.prot.decode(t)` instead of `t.decode()`

Issue #153, opened 2018-09-20
Create private compare protocol


bendecoste:
child of #106

PR #201, opened 2018-10-01
wrote prot.equal, prot.where & int32.neg methods

bendecoste:
needed for private compare and may be useful in the future

related #153 

I put this in it's own PR to keep the sizes down

### DIFF
--- a/tensorflow_encrypted/protocol/pond.py
+++ b/tensorflow_encrypted/protocol/pond.py
@@ -427,6 +427,24 @@ class Pond(Protocol):
         return self.sum(x, axis, keepdims)
 
     @memoize
+    def where(self, x):
+        with tf.name_scope('where'):
+            x_on_0, x_on_1 = x.unwrapped
+
+            with tf.device(self.server_0.device_name):
+                z_on_0 = tf.where(x_on_0.value)
+
+            with tf.device(self.server_1.device_name):
+                z_on_1 = tf.where(x_on_1.value)
+
+            return PondPublicTensor(self, self.tensor_factory.Tensor.from_native(z_on_0), self.tensor_factory.Tensor.from_native(z_on_1), x.is_scaled)
+
+    @memoize
+    def equal(self, x, y):
+        x, y = self.lift(x, y)
+        return self.dispatch('equal', x, y)
+
+    @memoize
     def sub(self, x, y):
         x, y = self.lift(x, y)
         return self.dispatch('sub', x, y)
@@ -878,7 +896,9 @@ class PondPublicTensor(PondTensor):
     ) -> None:
         assert isinstance(value_on_0, AbstractTensor), type(value_on_0)
         assert isinstance(value_on_1, AbstractTensor), type(value_on_1)
-        assert value_on_0.shape == value_on_1.shape
+
+        # as list is performed to compare partially unknown shapes (e.g. (?, 1) & (?, 1))
+        assert value_on_0.shape.as_list() == value_on_1.shape.as_list()
 
         super(PondPublicTensor, self).__init__(prot, is_scaled)
         self.value_on_0 = value_on_0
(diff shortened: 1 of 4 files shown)


tensorflow_encrypted/protocol/pond.py
```diff
@@ -878,7 +896,9 @@ def __init__(
     ) -> None:
         assert isinstance(value_on_0, AbstractTensor), type(value_on_0)
         assert isinstance(value_on_1, AbstractTensor), type(value_on_1)
-        assert value_on_0.shape == value_on_1.shape
+
+        # as list is performed to compare partially unknown shapes (e.g. (?, 1) & (?, 1))
+        assert value_on_0.shape.as_list() == value_on_1.shape.as_list()
```
justin1121: pretty sure @mortendahl  had some reasons for the shape needing to be always defined

bendecoste: uh oh .. it is never known out of equal

bendecoste:
> I think this will need to be completed for `Int100Tensor`. And I think there might be problems cause by the shape not being defined..

Since it is currently being completed for `PrivateCompare` can this be deferred or will it cause too many problems immediately ?

tensorflow_encrypted/protocol/pond.py
justin1121: Yeah that makes sense. I think the potential problems will be caused by `Int100Tensor` not 100% sure... Maybe we just need to deal with it if we need to use `tf.equal`.

justin1121:
Hmmm up to you I suppose! It will need to be done for `Int100Tensor` as soon as @mortendahl has the binarize stuff completed for Int100.

bendecoste:
> Hmmm up to you I suppose! It will need to be done for `Int100Tensor` as soon as @mortendahl has the binarize stuff completed for Int100.

Okay let's wait to see what @mortendahl says about that + the shape assert

justin1121:
Ignore the bits about `Int100Tensor`. You might only need this for `PrimeTensor`?

tensorflow_encrypted/protocol/pond.py
mortendahl: yeah we need the shape to always be known, ie. `x.shape.is_fully_defined()` must be true. what do you mean by "it is never known out of equal"?

tensorflow_encrypted/protocol/pond.py
```diff
@@ -1563,6 +1583,26 @@ def _sum_masked(prot: Pond,
...
+    y_on_0, y_on_1 = y.unwrapped
+
+    with tf.name_scope('equal'):
+        with tf.device(prot.server_0.device_name):
+            z_on_0 = tf.equal(x_on_0.value, y_on_0.value)
+
+        with tf.device(prot.server_1.device_name):
+            z_on_1 = tf.equal(x_on_1.value, y_on_1.value)
+
+    return PondPublicTensor(prot, prot.tensor_factory.Tensor.from_native(z_on_0), prot.tensor_factory.Tensor.from_native(z_on_1), x.is_scaled)
```
mortendahl: you have to do the whole `from_native` thing on the two servers, otherwise the underlying operations (eg in case the crt tensor is used) are not pinned to them correctly

mortendahl: actually, these should be forwarded to `where` calls on the tensors, which will know how to correctly interpret their values -- the above won't work for crt tensors

tensorflow_encrypted/tensor/int32.py
```diff
@@ -58,6 +58,12 @@ def __mul__(self, other: Any) -> 'Int32Tensor':
     def __mod__(self, k: int) -> 'Int32Tensor':
         return self.mod(k)
 
+    def __neg__(self) -> 'Int32Tensor':
+        return self.mul(-1)
+
+    def negative(self) -> 'Int32Tensor':
+        return self.mul(Int32Tensor(tf.constant(-1)))
```
mortendahl: can you just pass `-1` here as well, as done above?

tensorflow_encrypted/protocol/pond.py
```diff
@@ -426,6 +426,24 @@ def sum(self, x, axis=None, keepdims=None):
     def reduce_sum(self, x, axis=None, keepdims=None):
         return self.sum(x, axis, keepdims)
 
+    @memoize
+    def where(self, x):
```
mortendahl: maybe add a type here, as it will certainly fail for anything the public tensors :D

tensorflow_encrypted/protocol/pond.py
```diff
@@ -426,6 +426,24 @@ def sum(self, x, axis=None, keepdims=None):
...
+        with tf.name_scope('where'):
+            x_on_0, x_on_1 = x.unwrapped
+
+            with tf.device(self.server_0.device_name):
+                z_on_0 = tf.where(x_on_0.value)
+
+            with tf.device(self.server_1.device_name):
+                z_on_1 = tf.where(x_on_1.value)
+
+            return PondPublicTensor(self, self.tensor_factory.Tensor.from_native(z_on_0), self.tensor_factory.Tensor.from_native(z_on_1), x.is_scaled)
```
mortendahl: see comments for `_equal_public_public` -- you'll need to forward calls to `where` which should also take care of calling `from_native`

tensorflow_encrypted/protocol/pond.py
bendecoste: sorry, I mean out of `where` .. e.g.
```python
a = tf.constant(np.array([1,0,1,0]))
eq = tf.equal(a, 0)
w = tf.where(eq)
print(w)
```
will print

```
Tensor("Where:0", shape=(?, 1), dtype=int64)
```

mortendahl: [the doc](https://www.tensorflow.org/api_docs/python/tf/where) says "Returns:
A Tensor with the same type and shape as x, y if they are non-None", so what if you do `tf.reshape(w, a.shape)` afterwards?

bendecoste:
closing this as it is all in the private compare PR

PR #203, opened 2018-10-01
Odd Implicit Maybe Prime Modulus Tensor

justin1121:
First go at a new tensor type that encapsulates some logic for wrapping on an odd modulus.

### DIFF
--- a/tensorflow_encrypted/tensor/odd_implicit.py
+++ b/tensorflow_encrypted/tensor/odd_implicit.py
@@ -10,10 +10,10 @@ from .tensor import AbstractTensor
 
 class OddImplicitTensor(AbstractTensor):
     def __init__(self, value: Union[np.ndarray, tf.Tensor], dtype=tf.int32) -> None:
-        if dtype is not tf.int32 and dtype is not tf.int64:
+        if dtype not in [tf.int32, tf.int64]:
             raise Exception("Only int32 and int64 dtypes are allowed for the odd implicit tensor")
         self.value = value
-        self.dtype = tf.int32
+        self.dtype = dtype
 
     @staticmethod
     def from_native(value: Union[np.ndarray, tf.Tensor], dtype=tf.int32) -> 'OddImplicitTensor':
@@ -23,7 +23,7 @@ class OddImplicitTensor(AbstractTensor):
     @staticmethod
     def sample_uniform(shape: Union[Tuple[int, ...], tf.TensorShape], dtype=tf.int32) -> 'OddImplicitTensor':
         return OddImplicitTensor(tf.random_uniform(shape=shape, dtype=dtype,
-                                                   minval=dtype.min, maxval=dtype.max), dtype)
+                                                   minval=dtype.min + 1, maxval=dtype.max), dtype)
 
     @staticmethod
     def sample_bounded(shape: List[int], bitlength: int) -> 'OddImplicitTensor':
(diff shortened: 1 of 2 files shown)


tensorflow_encrypted/tensor/odd_implicit.py
```diff
@@ -0,0 +1,126 @@
...
+        z = x.value + y.value
+
+        def func(val):
+            return tf.cond(val >= self.modulus, lambda: val + 1, lambda: val)
+
+        ret = tf.map_fn(func, z)
+
+        return OddImplicitTensor(ret, self.modulus, dtype=self.dtype)
+
+    def sub(self, other: Union['OddImplicitTensor', int]) -> 'OddImplicitTensor':
```
justin1121: @mortendahl feel like this sub bit isn't right... we still need to compensate if we underflow?

mortendahl:
for reference: https://stackoverflow.com/questions/199333/how-to-detect-integer-overflow

tensorflow_encrypted/tensor/odd_implicit.py
```diff
@@ -0,0 +1,137 @@
...
+
+            negative = lambda: tf.cond(x < tf.int32.min - y, lambda: x + y - 1, lambda: x + y)
+
+            maybe_negative = lambda: tf.cond(y < 0, negative, lambda: x + y)
+
+            positive = lambda: tf.cond(x > tf.int32.max - y, lambda: x + y + 1, lambda: x + y)
+
+            return tf.cond(y > 0, positive, maybe_negative)
+
+        ret = tf.map_fn(func, (x.value, y.value), dtype=self.dtype)
```
mortendahl: fyi you could alternatively express this using `tf.where` (not sure which approach is fastest):
```
z = x + y

# correct for overflow where needed
z = z + tf.where(
    tf.logical_and(y > 0, x > tf.int32.max - y),
    tf.ones(z.shape, dtype=tf.int32),
    tf.zeros(z.shape, dtype=tf.int32)
)

# correct for underflow where needed
z = z - tf.where(
    tf.logical_and(y < 0, x < tf.int32.min - y),
    tf.ones(z.shape, dtype=tf.int32),
    tf.zeros(z.shape, dtype=tf.int32)
)
```

tensorflow_encrypted/tensor/odd_implicit.py
```diff
@@ -0,0 +1,137 @@
...
+from ..config import run
+from .tensor import AbstractTensor
+
+
+class OddImplicitTensor(AbstractTensor):
+    def __init__(self, value: Union[np.ndarray, tf.Tensor], dtype=tf.int32) -> None:
+        if dtype is not tf.int32 and dtype is not tf.int64:
+            raise Exception("Only int32 and int64 dtypes are allowed for the odd implicit tensor")
+        self.value = value
+        self.dtype = tf.int32
```
mortendahl: should be `self.dtype = dtype`

tensorflow_encrypted/tensor/odd_implicit.py
```diff
@@ -0,0 +1,137 @@
...
+import tensorflow as tf
+from typing import Union, Optional, List, Dict, Any, Tuple
+
+from ..config import run
+from .tensor import AbstractTensor
+
+
+class OddImplicitTensor(AbstractTensor):
+    def __init__(self, value: Union[np.ndarray, tf.Tensor], dtype=tf.int32) -> None:
+        if dtype is not tf.int32 and dtype is not tf.int64:
```
mortendahl: could alternatively be `if dtype not in [tf.int32, tf.int64]:`

tensorflow_encrypted/tensor/odd_implicit.py
```diff
@@ -0,0 +1,137 @@
...
+
+    @staticmethod
+    def from_native(value: Union[np.ndarray, tf.Tensor], dtype=tf.int32) -> 'OddImplicitTensor':
+        assert isinstance(value, (np.ndarray, tf.Tensor)), type(value)
+        return OddImplicitTensor(value, dtype=dtype)
+
+    @staticmethod
+    def sample_uniform(shape: Union[Tuple[int, ...], tf.TensorShape], dtype=tf.int32) -> 'OddImplicitTensor':
+        return OddImplicitTensor(tf.random_uniform(shape=shape, dtype=dtype,
+                                                   minval=dtype.min, maxval=dtype.max), dtype)
```
mortendahl: should be `minval=dtype.min+1`, as that's the min we're implicitly using

tensorflow_encrypted/tensor/odd_implicit.py
mortendahl: is this still a concern?

tensorflow_encrypted/tensor/odd_implicit.py
```diff
@@ -0,0 +1,137 @@
...
+        return OddImplicitTensor(tf.transpose(self.value, perm), self.dtype)
+
+    def strided_slice(self, args: Any, kwargs: Any) -> 'OddImplicitTensor':
+        return OddImplicitTensor(tf.strided_slice(self.value, *args, **kwargs), self.dtype)
+
+    def reshape(self, axes: Union[tf.Tensor, List[int]]) -> 'OddImplicitTensor':
+        return OddImplicitTensor(tf.reshape(self.value, axes), self.dtype)
+
+
+def _lift(x: Union['OddImplicitTensor', int], dtype) -> 'OddImplicitTensor':
```
mortendahl: you could alternatively move this guy into the class and use the `self` parameter instead of passing `dtype`

tensorflow_encrypted/tensor/odd_implicit.py
justin1121: I'm pretty sure its right, wouldn't mind a second pair of eyes though!

PR #204, opened 2018-10-01
Extract average prediction time from trace file

yanndupis:
Hey, 

This script extract prediction time from the different trace files then compute the average and median. I'd like to add this script to master, then build a new docker image in order to compute the average prediction time directly in GCP instead of downloading tracing files locally. 

Thank you.

### DIFF
--- a/tools/get_avg_prediction_time.py
+++ b/tools/get_avg_prediction_time.py
@@ -1,0 +1,51 @@
+import glob
+import json
+import sys
+import argparse
+
+import numpy as np
+
+parser = argparse.ArgumentParser(description='Extract the average prediction\
+                                    time with the specified trace directory')
+parser.add_argument('trace_dir', type=str, help='Specify the trace directory')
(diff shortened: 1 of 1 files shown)


tools/get_avg_prediction_time.py
```diff
@@ -0,0 +1,51 @@
...
+
+    min_ts = timestamps[0]
+    max_ts = timestamps[-1]
+    return max_ts[1] - min_ts[0]
+
+
+durations = []
+
+for filename in glob.glob(trace_path):
+    # Exclude init0.ctr and prediction0.ctr as they don't reflect the real prediction time.
```
mortendahl: ideally take a list of files as input instead since the names `init0.ctr` and `prediction0.ctr` are arbitrary

yanndupis: Agreed, I will change this in a separate pr. Thank you.

PR #205, opened 2018-10-04
memoize decode to avoid adding new nodes to graph whenever called

mortendahl:


tensorflow_encrypted/session.py
```diff
@@ -8,6 +8,7 @@
 from tensorflow.python import debug as tf_debug
 
 from .config import Config, RemoteConfig, get_config
+# from .protocol.pond import PondPublicTensor  # TODO[Morten] can't do this because of circular import, should be fixed
```
morgangiraud: I agree but currently, It feels like tensor/protocol/config/session should be one file to fix this. 
Since we are missing headers file, we are running into those circular dependencies because of types most of the time.
Depending on what should be done next. I might take a look at it

morgangiraud: The problem is coming from the `eval` function at `protocol` level and `tensor` level.
both take as input a session. If we want to type it I don't see a way to avoid circular dependency cleanly.

PR #207, opened 2018-10-05
Implement a default protocol

morgangiraud:
Connect #175 

A first system to implement a default protocol, revealing its public function to the `tfe` module.

Tests are incoming. each one is already working, but since this changes introduce a global state to the module, they break when run in parallel. I need to find a way to make them self-contained.

### DIFF
--- a/examples/matmul.py
+++ b/examples/matmul.py
@@ -22,15 +22,15 @@ import tensorflow_encrypted as tfe
 # Setting your custom configuration
 # tfe.set_config(config)
 
-# The default protocol is called pond, it's a 3 party computation protocol
-# But you can use your an other one like securreNN!
+# TFE sets a default cryptographic protocol, called Pond
+# It's a 3 party computation protocol
+# But you can use your another one!
 # tfe.set_protocol(tfe.protocol.SecureNN())
 
-w = tfe.define_private_variable(np.zeros((10, 10)))
-# x = tfe.define_private_variable(np.zeros((1,100)))
+w = tfe.define_private_variable(np.ones((10, 10)))
 
 y = w
-for _ in range(5):
+for _ in range(2):
     y = y.matmul(y)
 
 with tfe.Session() as sess:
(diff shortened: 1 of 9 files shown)


examples/matmul.py
```diff
@@ -22,15 +22,17 @@
 # Setting your custom configuration
 # tfe.set_config(config)
 
-with tfe.protocol.Pond() as prot:
+# The default protocol is called pond, it's a 3 party computation protocol
```
mortendahl: grammatically these two lines could be cleaned up a bit :P

tensorflow_encrypted/__init__.py
```diff
@@ -1,27 +1,72 @@
...
-from .protocol import global_caches_updator
+from .protocol import global_caches_updator, Pond, get_protocol
 from .player import player
 from . import io
 from . import protocol
 from . import layers
 from . import convert
 
 
+def get_protocol_public_func(prot: protocol.Protocol) -> list:
```
mortendahl: is this called somewhere or pure black magic?

tensorflow_encrypted/protocol/__init__.py
```diff
@@ -1,11 +1,29 @@
...
-# from unencrypted_fixedpoint import UnencryptedFixedpoint
 from .pond import Pond, TFEVariable, TFETensor
 from .securenn import SecureNN
 
-from .protocol import Protocol, global_caches_updator, memoize
+from .protocol import Protocol, global_caches_updator, memoize, set_protocol, get_protocol
+
+
+def get_all_funcs() -> list:
+    all_prot_methods = []
```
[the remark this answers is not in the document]
morgangiraud: Done for the list. 

Currently, I need to instantiate as some of the functions are not pure functions and rely on some states inside the classes.

mortendahl:
@morgangiraud ready to merge this guy?

PR #212, opened 2018-10-09
Private compare suggestions

mortendahl:


tensorflow_encrypted/protocol/pond.py
```diff
@@ -1007,48 +982,34 @@ def __init__(
...
         return (self.share0, self.share1)
 
     def reveal(self) -> PondPublicTensor:
         return self.prot.reveal(self)
 
-    @staticmethod
-    def zero(prot: Pond, shape: List[int]) -> 'PondPrivateTensor':
-        zero = np.zeros(shape=shape)
-        return prot.define_private_variable(zero)
+    # @staticmethod
```
bendecoste: either implement or delete?

tensorflow_encrypted/protocol/pond.py
```diff
@@ -905,16 +918,8 @@ def expand_dims(self):
     def reshape(self, shape: List[int]) -> 'PondTensor':
         return self.prot.reshape(self, shape)
 
-    def broadcast(self, to: List[int]) -> 'PondTensor':
```
bendecoste: any reason to delete these? just b/c they aren't implemented via dispatch?

bendecoste: same question re where & gather_nd

mortendahl: only because the code seemed a bit too specific, making assumptions about the internals of the backing tensors (would fail for int100 for instance)

tensorflow_encrypted/protocol/pond.py
mortendahl: let's delete!

Issue #123, opened 2018-09-18
Guides and API Documentation


ianlivingstone:
### Background

Today, we have a few examples that we've put together along with a short getting started guide. It's great for the stage of the project.

However, as we build towards a beta or more stable release we'll need to expand upon these guides and properly document the API.

This issue is a discussion for the type of documentation we require, where it should live, and the targeted audience.

### TODO

- [x] Establish the structure and process for deploying documentation (via readthedocs.org.)
- [x] Update README with link to docs and other getting started components
- [x] Write a getting started guide that uses google colab (public training, private prediction, and loading a pretrained model)
- [x] Update tf-encrypted on PyPI so we can get access to the latest API in Google Colab.
- [x] Write a overview of our technology (can link off to blog post or paper) with architecture diagram
- [ ] Add link to paper to the README and Docs Landing Page
- Flush out the API documentation with examples, function signature, and other information.
  - [ ] session
  - [ ] config
  - [ ] define_pub/private_variable
  - [ ] pondtensors
  - protocols
     - [ ] pond
     - [ ] secureNN
  - [ ] input
  - [ ] output

ianlivingstone:
I setup time for the dropoutlabs team to discuss this issue.

bendecoste:
I took some time to play around with how writing and using readthedocs would actually work.  I think this is a good option going forward.  The highlights are

- We can use inline docs, but you need to manually link to them (more on this below)
- It is easy to write guides and have API docs in the same place
- Markdown is supported (though they use `.rst` by default, which is similar but more powerful)

~~~~~

> We can use inline docs, but you need to manually link to them (more on this below)

Re inline docs.  As far as I can tell you do indeed need to manually link to your inline documentation, but you can do this on a module level, so in general it is more or less automatic.  For example, 

```
.. automodule:: tensorflow_encrypted.protocol
  :members:
```

will generate all of the docs that we have written for .protocol

![image](https://user-images.githubusercontent.com/1411145/46676859-c0dd9180-cbb7-11e8-931d-77c68f8825c4.png)

> It is easy to write guides and have API docs in the same place

Again quite easy, you can make entire pages devoted to API as done above, it is also easy to intertwine or write a page that is just a guide

> Markdown is supported (though they use `.rst` by default, which is similar but more powerful)

Here is a simple example

![image](https://user-images.githubusercontent.com/1411145/46676937-f3878a00-cbb7-11e8-9d79-f77107b20b20.png)

renders

[...]

PR #213, opened 2018-10-09
Skeleton structure for docs

bendecoste:
related #123

This sets up Sphynx which will automatically be used by readthedocs.


docs/source/conf.py
```diff
@@ -0,0 +1,187 @@
...
+# -- Project information -----------------------------------------------------
+
+project = 'tf-encrypted'
+copyright = '2018, Dropout Labs'
+author = 'Dropout Labs'
+
+# The short X.Y version
+version = ''
+# The full version, including alpha/beta/rc tags
+release = ''
```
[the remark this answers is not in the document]
bendecoste: I don't believe so.  setup isn't in any tfe module.  I can see about adding a version property onto tfe then we could require that

docs/source/conf.py
```diff
@@ -0,0 +1,187 @@
...
+     [author], 1)
+]
+
+
+# -- Options for Texinfo output ----------------------------------------------
+
+# Grouping the document tree into Texinfo files. List of tuples
+# (source start file, target name, title, author,
+#  dir menu entry, description, category)
+texinfo_documents = [
```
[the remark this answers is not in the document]
bendecoste: Added a one liner .. this is for latext output so not super high priority.

docs/source/conf.py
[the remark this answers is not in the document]
ianlivingstone: `__version__` seems to be the standard on the main module thne it gets sourced into `setup.py`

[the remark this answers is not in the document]
bendecoste: yeah .. ideally we could source the version from the git tag .. not really in scope of this PR imo

docs/source/api/config/config.rst
```diff
@@ -0,0 +1,22 @@
+`Config`
+============
+
+| Config determines how a session should run in tf-encrypted.
+| There are two primary ways in which config can be used:
+.. TODO: How do we link these to the real thing?
```
ianlivingstone: You mean the actual class api?

docs/source/api/config/config.rst
```diff
@@ -0,0 +1,22 @@
...
+.. py:class:: tensorflow_encrypted.config.LocalConfig
+and
+
+.. py:class:: tensorflow_encrypted.config.RemoteConfig
+As the name implies, `LocalConfig` is used to create a
+local session. This is useful for quick debugging and prototyping.
```
ianlivingstone: Perhaps instead of local session we should define what that means: to run an entire session locally on your compute where as RemoteConfig is used to specify how to run a session in a distributed setting (e.g. using multiple machines).

docs/source/api/index.rst
```diff
@@ -0,0 +1,24 @@
...
+
+The top tfe module.
+
+.. code-block:: python
+
+    import tensorflow_encrypted as tfe
+
+
+.. automodule:: tensorflow_encrypted
+  :members: set_protocol, set_config
```
ianlivingstone: Do we have to specify every function to be documented?

docs/source/api/layers/index.rst
```diff
@@ -0,0 +1,18 @@
+`layers`
+=========
+
+Layers implements ready to use `neural network` layers that come with crypto for free.
```
ianlivingstone: Is there another way we can write this other than with crypto for free? A user may not know what "crypto for free" means? 

cc @mortendahl

docs/source/api/protocol/index.rst
```diff
@@ -0,0 +1,27 @@
+`protocol`
+===========
+
+In `tf-encrypted`, a protocol represents a certain type of cryptographic protocol
+to achieve security.
+
+The goal is that you can easily play with or use different cryptographic methods
```
ianlivingstone: We should probably document why you'd want to change protocols? e.g. for benchmarking purposes etc

cc @mortendahl

docs/source/api/protocol/pond.rst
```diff
@@ -0,0 +1,13 @@
+`Pond`
+=======
+
+Pond is an implementation of the SPDZ algorithm for MPC.
```
ianlivingstone: Can we link off to definitions of SPDZ and MPC?

docs/source/api/protocol/securenn.rst
```diff
@@ -0,0 +1,9 @@
+`SecureNN`
+============
+
+SecureNN is an extension of the `Pond` protocol.  It is an implementation from the `SecureNN paper`_.
```
ianlivingstone: Perhaps we should specify the advantages it has over Pond? (e.g. why use this)

docs/source/api/session.rst
```diff
@@ -0,0 +1,16 @@
...
+Session is an extension of `tf.Session` that lets the graph run in a secure manner.
+
+You get and use a session in the same manner that you are used to with Tensorflow::
+
+    import tensorflow_encrypted as tfe
+
+    with tfe.Session() as sess:
+      # sess.run like normal
+
+TODO -- More coming soon.
```
ianlivingstone: Is this a todo we'll resolve in this PR?

docs/source/api/session.rst
```diff
@@ -0,0 +1,16 @@
+`Session`
+==========
+
+Session is an extension of `tf.Session` that lets the graph run in a secure manner.
```
ianlivingstone: Is it an extension or a wrapper around tfe.Session? We could just document that it wraps tf.Session and handles configuring the session for secure computation with the chosen protocol.

docs/source/api/index.rst
bendecoste: no . this page is not in use currently

PR #214, opened 2018-10-09
Private annotation

### LINKED ISSUE #175: Make the API closer to TF

morgangiraud:
closes #175 
Look at the two files side by side `tf-logreg` and `tfe-logreg`.

Notice that now I print the accuracy "securely"


tensorflow_encrypted/protocol/pond.py
```diff
@@ -187,21 +186,25 @@ def define_private_variable(
 
     def define_public_input(
         self,
-        provider: InputProvider,
+        player: Union[str, Player],
+        data_fn: Callable,
```
mortendahl: Think it's worth high lighting that it's a Callable that returns either a single tf.Tensor or list of tf.Tensors

tensorflow_encrypted/protocol/pond.py
```diff
@@ -262,11 +269,15 @@ def helper(v: tf.Tensor):
 
     def define_output(
         self,
+        player: Union[str, Player],
         xs: Union['PondPrivateTensor', List['PondPrivateTensor']],
-        receiver: OutputReceiver,
+        post_fn: Callable,
```
mortendahl: could we call it `receiver_fn` instead? think that fits better with standard terminology

mortendahl: less important with a stronger type for the callable here, but could still be done

tensorflow_encrypted/protocol/pond.py
```diff
@@ -218,14 +221,18 @@ def helper(v: tf.Tensor) -> 'PondPublicTensor':
 
     def define_private_input(
         self,
-        provider: InputProvider,
+        player: Union[str, Player],
+        data_fn: Callable,
```
mortendahl: could we call it `inputter_fn` instead? think that's closer to the standard mpc terminology of "input provider"

mortendahl: alternatively `provider_fn` but maybe that's less clear

mortendahl: stronger typing also important here (users might try to pass functions returning numpy arrays otherwise)

tensorflow_encrypted/protocol/pond.py
mortendahl: see naming comments elsewhere, same applies here

examples/federated-average/run.py
```diff
@@ -20,36 +20,34 @@
...
 server0 = config.get_player('server0')
 server1 = config.get_player('server1')
 crypto_producer = config.get_player('crypto_producer')
 
 with tfe.protocol.Pond(server0, server1, crypto_producer) as prot:
 
     # get input from inputters as private values
-    inputs = [prot.define_private_input(inputter)[0] for inputter in inputters]
+    inputs = [
+        prot.define_private_input(config.get_player(name), provide_input)[0] for name in player_names_inputter
```
mortendahl: you can avoid the `[0]` here by returning a `tf.Tensor` from `provide_input` instead of a list

examples/federated-average/run.py
```diff
@@ -20,36 +20,34 @@
...
+    ]
 
     # sum all inputs and divide by count
     result = reduce(lambda x, y: x + y, inputs) * (1 / len(inputs))
 
     # send result to receiver
-    result_op = prot.define_output([result], result_receiver)
+    result_op = prot.define_output(config.get_player('result_receiver'), [result], receive_output)
```
mortendahl: is it possible to avoid call to `config.get_player` (and pass string instead)?

examples/mnist/run.py
```diff
@@ -131,22 +137,22 @@ def receive_output(self, likelihoods: tf.Tensor) -> tf.Operation:
...
+    w0, b0, w1, b1 = params
     layer0 = prot.matmul(x, w0) + b0
     layer1 = prot.sigmoid(layer0 * 0.1)  # input normalized to avoid large values
-    layer2 = prot.matmul(layer1, w1) + b1
-    prediction = layer2
+    logits = prot.matmul(layer1, w1) + b1
 
     # send prediction output back to client
-    prediction_op = prot.define_output(prediction, prediction_client)
+    prediction_op = prot.define_output(prediction_client.player, [logits], prediction_client.receive_output)
```
mortendahl: should be `logits` instead of `[logits]`, no?

tensorflow_encrypted/protocol/pond.py
```diff
@@ -262,11 +273,16 @@ def helper(v: tf.Tensor):
 
     def define_output(
         self,
+        player: Union[str, Player],
         xs: Union['PondPrivateTensor', List['PondPrivateTensor']],
-        receiver: OutputReceiver,
+        post_fn: Callable[..., Any],
```
mortendahl: this guy is a `Callable[tf.Tensor, ...]`if i understand that syntax correct

morgangiraud: Not really, it can take any number of parameters:

`def post_fn(*args: tf.Tensor) -> Any`

But I didn't find any way to type this `*args` in `Callable` definition. I would need something like `Callable[...:tf.Tensor, Any]` or `Callable[StarArg: tf.Tensor, Any]` but it's not working

Also, I just noticed I didn't rename it, what name would you prefer? `outputter_fn`?

examples/mnist/run.py
morgangiraud: Both are allowed. I can change it if you prefer?

examples/federated-average/run.py
morgangiraud: Yes

tensorflow_encrypted/protocol/pond.py
mortendahl: yeah, `outputter_fn` works 👍

PR #228, opened 2018-10-11
Implements reduce_max operation for SecureNN

mortendahl:


tensorflow_encrypted/protocol/securenn.py
```diff
@@ -133,8 +139,29 @@ def relu(self, x):
...
-    def max_pool(self, x):
-        raise NotImplementedError
+    @memoize
+    def maximum(self, x, y):
+        with tf.name_scope('maximum'):
+            indices_of_maximum = self.greater(x, y)
+            return self.select(indices_of_maximum, y, x)
+
+    @memoize
+    def reduce_max(self, x, axis=0):
```
morgangiraud: side note: I understand this is not a PR about `max_pooling` but `tf.max_pool`get the max on up to a `4d-tensor`. It will need to implement multiple reduction axis support unless you intend to use some kind of reshape?

PR #233, opened 2018-10-12
basic int64 support 

### LINKED ISSUE #79: int64 native support in TF
Using int64 natively in TF, would allow to gain substantial speed up by leveraging the ring it creates.

We need to investigate how much work should be done to use int64 operations in TF.

related https://github.com/tensorflow/tensorflow/pull/21253

mortendahl:
Adds support for int64 backing tensors and sets it as default when TensorFlow supports it

Closes #79

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ all: test
 # ###############################################
 DOCKER_REQUIRED_VERSION=18.
 PYTHON_REQUIRED_VERSION=3.5.
-TENSORFLOW_REQUIRED_VERSION=1.11
+TENSORFLOW_REQUIRED_VERSION=1.10
 SHELL := /bin/bash
 
 CURRENT_DIR=$(shell pwd)
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ all: test
 # ###############################################
 DOCKER_REQUIRED_VERSION=18.
 PYTHON_REQUIRED_VERSION=3.5.
-TENSORFLOW_REQUIRED_VERSION=1.10
+TENSORFLOW_REQUIRED_VERSION=1.11
 SHELL := /bin/bash
 
 CURRENT_DIR=$(shell pwd)
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ bootstrap: pythoncheck pipcheck
 #
 # Rules for running our tests and for running various different linters
 # ###############################################
-test: lint pythoncheck tensorflowcheck
+test: lint pythoncheck
 	python examples/convert.py
 	python examples/inputs.py
 	python examples/int32.py
@@ -69,10 +69,10 @@ test: lint pythoncheck tensorflowcheck
 	python examples/federated-average/run.py
 	python -m unittest discover
 
-lint: pythoncheck tensorflowcheck
+lint: pythoncheck
 	flake8
 
-typecheck: pythoncheck tensorflowcheck
+typecheck: pythoncheck
 	MYPYPATH=$(CURRENT_DIR):$(CURRENT_DIR)/stubs mypy tensorflow_encrypted
 
 
--- a/examples/matmul.py
+++ b/examples/matmul.py
@@ -42,4 +42,4 @@ with tfe.Session() as sess:
     actual = sess.run(y.reveal(), tag='reveal')
 
     expected = b
-    np.testing.assert_allclose(actual, expected, atol=.1)
+    np.testing.assert_allclose(actual, expected)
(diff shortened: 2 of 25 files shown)


Makefile
```diff
@@ -58,7 +58,7 @@ bootstrap: pythoncheck pipcheck
 #
 # Rules for running our tests and for running various different linters
 # ###############################################
-test: lint pythoncheck tensorflowcheck
+test: lint pythoncheck
```
mortendahl: @ianlivingstone what do you think about leave `tensorflowcheck` out of these three? having them got in the way when I was testing tf-int64 but maybe there's a better way that allows us to keep them?

mortendahl:
how should we deal with the need for a custom build of TF for int64:
- can we somehow detect which build is available and choose different defaults?
- how do we make sure to run tests for both int100 and int64?

mortendahl:
hacky test added in Pond for detecting when int64 is available, falling back to int100 when not

tensorflow_encrypted/protocol/securenn.py
```diff
@@ -40,8 +40,11 @@ def __init__(
             **kwargs
         )
         self.server_2 = server_2
-        self.prime_factory = prime_factory or PrimeFactory(107)
+
+        self.prime_factory = prime_factory or PrimeFactory(107, native_type=self.tensor_factory.native_type)
```
jvmancuso: I've been thinking about this part of using `kwarg=None` in combination with `x = kwarg or default_value`. I think it's less readable and I think we should change these kinds of antipatterns to: 
```
if kwarg is not None:
     x = kwarg
else:
     ...
```

tensorflow_encrypted/protocol/securenn.py
```diff
@@ -73,11 +76,13 @@ def bitwise_xor(self, x: PondTensor, y: PondTensor) -> PondTensor:
     @memoize
     def msb(self, x: PondTensor) -> PondTensor:
         # NOTE when the modulus is odd then msb reduces to lsb via x -> 2*x
-        if x.backing_dtype.modulus % 2 != 1:
-            # NOTE: this is currently only for use with an odd-modulus CRTTensor
-            #       NativeTensor will use an even modulus and will require share_convert
-            raise Exception('SecureNN protocol assumes a ring of odd cardinality, ' +
-                            'but it was initialized with an even one.')
+
```
jvmancuso: should we remove this now or wait until share convert is fixed?

tests/test_int64_tensor.py
```diff
@@ -3,23 +3,21 @@
...
-        config = tfe.LocalConfig([
-            'server0',
-            'server1',
-            'crypto_producer'
-        ])
-
-        with tfe.protocol.Pond(*config.get_players('server0, server1, crypto_producer'),
-                               tensor_factory=Int64Factory(), use_noninteractive_truncation=True,
-                               verify_precision=False) as prot:
+
```
jvmancuso: could use more comprehensive testing, perhaps we can make a separate issue about it

tests/test_matmul.py
```diff
@@ -83,7 +75,7 @@ def test_big_middle_matmul(self) -> None:
             sess.run(tf.global_variables_initializer())
             out_tensorflow = sess.run(out)
 
-        np.testing.assert_array_almost_equal(out_pond, out_tensorflow, decimal=2)
+        np.testing.assert_allclose(out_pond, out_tensorflow, atol=.3)
```
jvmancuso: is this atol value a bit high?

examples/matmul.py
```diff
@@ -42,4 +42,4 @@
     actual = sess.run(y.reveal(), tag='reveal')
 
     expected = b
-    np.testing.assert_array_equal(actual, expected)
+    np.testing.assert_allclose(actual, expected, atol=.1)
```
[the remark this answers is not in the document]
mortendahl: i think it was the best we could do for non-interactive truncation and int64, but switched since interactive truncation can do better

tensorflow_encrypted/protocol/securenn.py
mortendahl: I agree it's not super easy to read

tensorflow_encrypted/protocol/securenn.py
mortendahl: I'll leave it commented out for now as we may still want to bench with exact relu and int64 despite getting wrong results

examples/matmul.py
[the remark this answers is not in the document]
jvmancuso: perhaps two different tests, one with interactive and one without?

tests/test_matmul.py
mortendahl: .1 is the best we can do due to int64

PR #257, opened 2018-10-22
Naming

mortendahl:
Small updates, mainly consistent naming of players


ianlivingstone:
I believe we will need to update the naming inside our benchmarking environment. cc @yanndupis

bendecoste:
Is there any difference between this branch and #256  ?

mortendahl:
@bendecoste good catch! no difference, but github was acting up earlier today. i've closed the other issue to keep @ianlivingstone's comment alive

ianlivingstone:
A comment should be added to the changelog that we've modified those variables for specifying a node's role!

PR #259, opened 2018-10-23
Updates for the readme

mortendahl:


ianlivingstone:
The docs index and getting started guide should be updated as well :)

mortendahl:
@ianlivingstone done!

docs/getting_started.md
```diff
@@ -1,36 +1,35 @@
-# TensorFlow Encrypted
+# Getting Started with tf-encrypted
```
ianlivingstone: We can probably just remove this now that we have the main documentation. I don't think anything links here an maintaining two duplicate guides is a lot of work!

mortendahl: not sure I follow -- which part should be removed?

ianlivingstone: This file entirely as we have `docs/source/getting_started.rst` which is available at https://tf-encrypted.readthedocs.io/en/latest/usage/getting_started.html

PR #265, opened 2018-10-25
experimental secure random

justin1121:
Child of https://github.com/mortendahl/tf-encrypted/issues/273


secure_random/Makefile
```diff
@@ -0,0 +1,19 @@
+LIBSODIUM_VER_TAG=1.0.16
+
+TF_CFLAGS=$(shell python -c 'import tensorflow as tf; print(" ".join(tf.sysconfig.get_compile_flags()))' 2>/dev/null)
+TF_LFLAGS=$(shell python -c 'import tensorflow as tf; print(" ".join(tf.sysconfig.get_link_flags()))' 2>/dev/null)
+
+all: build-op
+
+build-lib-sodium:
+	# TODO CHECK IF SODIUM LIB EXISTS FIRST
+	git clone https://github.com/jedisct1/libsodium.git
```
ianlivingstone: Instead of cloning the entire repository, we can download the version from GitHub as listed on their releases page: https://github.com/jedisct1/libsodium/releases

ianlivingstone: We should probably put this into a `build/` folder that is git ignored. I'd also suggest that these make commands be a part of the larger Makefile as it's going to all be integrated together as a part of the same build pipeline.

secure_random/Makefile
```diff
@@ -0,0 +1,19 @@
...
+	cd libsodium && git checkout $(LIBSODIUM_VER_TAG)
+	cd libsodium && ./autogen.sh && ./configure --disable-shared --enable-static \
+		--disable-debug --disable-dependency-tracking --with-pic
+	$(MAKE) -C libsodium
+	$(MAKE) -C libsodium install
+	rm -Rf libsodium
```
ianlivingstone: Deleting the source code should probably be a part of a `make clean` command

secure_random/Makefile
```diff
@@ -0,0 +1,19 @@
...
+
+build-op:
```
ianlivingstone: this should depend on the secure_random.cc file so it only builds when secure_random.cc changes. It could also depend on any file inside the `secure_random` file changing.

I'd also suggest we consider a different structure, something like `src/operations/secure_random` or similar so it's separated nicely from the rest of the project.

ianlivingstone: How o we specify the platform we want to build for?

secure_random/secure_random.cc
```diff
@@ -0,0 +1,90 @@
...
+    OP_REQUIRES_OK(context, context->allocate_output(0, shape, &output));
+    if (shape.num_elements() == 0) return;
+
+    if (sodium_init() < 0) {
+      return;
+    }
+  
+    int number_of_seeds = randombytes_SEEDBYTES / sizeof(int32);
+
+    int32 *array = (int *)malloc(sizeof(int32) * number_of_seeds);
```
ianlivingstone: We're mallocing here but I don't see us ever freeing the memory?

secure_random/secure_random.py
```diff
@@ -0,0 +1,4 @@
+import tensorflow as tf
```
ianlivingstone: ❤️ looking forward to seeing a pyunit test!

secure_random/Makefile
justin1121: > How do we specify the platform we want to build for?

Haven't worked on anything concretely for that yet but it will include building on the correct circle ci image and passed the `--plat-name` arg to the `python setup.py bdist_wheel` call like:

```
python setup.py bdist_wheel --plat-name macosx_10_7_x86_64 # for macos
```

operations/secure_random/secure_random.cc
```diff
@@ -0,0 +1,203 @@
...
+    auto range = static_cast<typename std::make_unsigned<T>::type>(hi) -
+                   static_cast<typename std::make_unsigned<T>::type>(lo);
+
+    typename std::make_unsigned<T>::type min = (1U + ~hi) % hi;
+
+    for (int i = 0; i < count_; ++i) {
+      auto unsign = static_cast<typename std::make_unsigned<T>::type>(buf_[i]);
+        while(unsign < min) {
+          // rejection sampling, get the next valid number in the stream
+          buf_[i] = GetNextValidData();
```
justin1121: things got a little crazy here with the rejection sampling. @mortendahl would love for you to give an in depth look into to this and make sure it makes sense! Let me know if you have any ?s.

operations/secure_random/Makefile
```diff
@@ -0,0 +1,38 @@
+LIBSODIUM_VER_TAG=1.0.16
```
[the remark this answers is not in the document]
justin1121: yeah i'm fine with that, was just nice to keep it separate during development

setup.py
```diff
@@ -7,6 +7,7 @@
     name="tf-encrypted",
     version="0.2.0",
     packages=setuptools.find_packages(),
+    package_data={'tf_encrypted': ['operations/secure_random/secure_random.so']},
```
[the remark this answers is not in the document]
justin1121: It'll just happen automatically once we have the circleci setup to build different version of wheels. I'll add some code for building different versions based on the tensorflow version though.

justin1121:
> How do we compile this for all the platforms we support and declare those supported platforms (OS + Architecture) in the setup.py for when we publish to pip? How will this get configured on Circle CI?

This isn't done in the setup.py. Its passed as an argument to `python setup.py bdist_wheel` and then we push those generated wheels to pypi. When you do a pip install, pip automatically detects your platform and downloads the correct build. As for circle ci we'll just have to configure a couple different workflows to do the builds and then push these releases to pypi.

> Once this lands we won't be able to release tf-encrypted until the Circle CI work is done unless we make the secure random op optional.

Currently disabled by default and not really exposed at all. Here: https://github.com/mortendahl/tf-encrypted/pull/265/files#diff-083ee9e1c17b69f21f464290a1fc8b0eR5. Someone would have to dig into the installed pip package code and change that to try and call the secure_random function. I think it'd be fine to continue to do builds. Figured we could include an exposed api for this once we figure out the tensor api/seeded randomness stuff.

tf_encrypted/operations/secure_random/secure_random.py
```diff
@@ -0,0 +1,47 @@
...
+    :param int minval: Minimum value to return, inclusive.
+    :param int maxval: Maximum value to return, exclusive.
+    :param dtype: Data type of the return random values. Either int32 or int64.
+    :param tf.Tensor seed: The seed to be used when generating the random numbers.
+    :param str name:
+
+    :rtype: tf.Tensor
+    """
+
+    secure_random_module = tf.load_op_library(shared_object)
```
mortendahl: could we do this once and for all, instead of each time `secure_random` is called?

tf_encrypted/operations/secure_random/secure_random.py
```diff
@@ -0,0 +1,47 @@
...
+import tf_encrypted as tfe
+from tensorflow.python.framework import ops
+from tensorflow.python.framework import dtypes
+import os
+
+dirname = os.path.dirname(tfe.__file__)
+shared_object = dirname + '/operations/secure_random/secure_random_module.so'
+
+
+def secure_random(shape, minval=0, maxval=None, dtype=tf.int32, seed=None, name=None):
```
mortendahl: I suggest we have three operations actually:
- one to generate seeds
- one to expand seeds into full tensors
- one to securely sample full tensors

the first two are strongly correlated, the third is unrelated (at least on the surface; it may use whatever underneath)

tf_encrypted/operations/secure_random/secure_random.py
```diff
@@ -0,0 +1,47 @@
...
+    if maxval is None:
+        raise ValueError("Must specify maxval for integer dtype %r" % dtype)
+
+    if seed is None:
+        # TODO random seed if not specified?
+        seed = [87654321, 87654321, 87654321, 87654321, 87654321, 87654321, 87654321, 87654321]
+    elif len(seed) != 8:
+        raise ValueError("Seed length must 8")
+
+    minval = ops.convert_to_tensor(minval, dtype=dtype, name="min")
```
mortendahl: is it possible to pass integers to the custom op?

tf_encrypted/tensor/int64.py
```diff
@@ -59,12 +61,12 @@ def native_type(self):
         return tf.int64
 
     def sample_uniform(self, shape: List[int]) -> 'Int64Tensor':
-        value = tf.random_uniform(shape=shape, dtype=self.native_type, minval=tf.int64.min, maxval=tf.int64.max)
+        value = random_func(shape=shape, dtype=self.native_type, minval=tf.int64.min, maxval=tf.int64.max)
```
mortendahl: let's stick with the `random_uniform` name! ie something like `tfe.random_uniform`

operations/secure_random/secure_random.cc
```diff
@@ -0,0 +1,203 @@
...
+    // Allocate output
+    Tensor* output;
+    OP_REQUIRES_OK(context, context->allocate_output(0, shape, &output));
+    if (shape.num_elements() == 0) return;
+
+    if (sodium_init() < 0) {
+      return;
+    }
+
+    int number_of_seeds = randombytes_SEEDBYTES / sizeof(int32);
```
mortendahl: where does `randombytes_SEEDBYTES` come from, and does it related to the `8` earlier?

operations/secure_random/secure_random.cc
```diff
@@ -0,0 +1,203 @@
...
+public:
+  Tensor *output = NULL;
+  const unsigned char * seeds = NULL;
+
+  Generator(Tensor* output, const unsigned char * seeds) : output(output), seeds(seeds) {
+    auto flat = output->flat<T>();
+
+    count_ = output->flat<T>().size();
+    bytes_count_ = count_ * sizeof(T);
+    buf_ = static_cast<T *>(malloc(bytes_count_));
```
mortendahl: is there anyway we can avoid this `malloc` and use what was already allocated by TF for the output (which might keep a memory pool internally)? or perhaps it doesn't work that way..?

operations/secure_random/secure_random.cc
```diff
@@ -0,0 +1,203 @@
...
+    .SetShapeFn(SecureRandomShape);
+
+// this function allows you to skip ahead in the chacha stream so you don't
+// have to allocate more memory than you need to, used in rejection sampling and
+// will be handy for parallelizing this operation
+void randombytes_buf_deterministic_ic(void * const buf, const size_t size, uint32_t count,
+                              const unsigned char seed[randombytes_SEEDBYTES])
+{
+    static const unsigned char nonce[crypto_stream_chacha20_ietf_NONCEBYTES] = {
+        'L', 'i', 'b', 's', 'o', 'd', 'i', 'u', 'm', 'D', 'R', 'G'
```
mortendahl: that doesn't look like a nonce to me :P was it used like this elsewhere?

operations/secure_random/secure_random.cc
mortendahl: want to add a reference to the method used?

operations/secure_random/secure_random.cc
```diff
@@ -0,0 +1,203 @@
...
+  ShapeHandle out;
+  TF_RETURN_IF_ERROR(context->MakeShapeFromShapeTensor(0, &out));
+  context->set_output(0, out);
+  return Status::OK();
+}
+
+REGISTER_OP("SecureRandom")
+    .Input("shape: T")
+    .Input("seed: Tseed")
+    .Input("minval: dtype")
```
mortendahl: why `dtype` btw?

Makefile
```diff
@@ -246,3 +264,36 @@ push: pypi-version-check
...
+# ###############################################
+
+$(LIBSODIUM_OUT):
+	wget -nc https://github.com/jedisct1/libsodium/archive/$(LIBSODIUM_VER_TAG).tar.gz
+	mkdir -p build
+	tar -xvf $(LIBSODIUM_VER_TAG).tar.gz -C build
+	cd $(LIBSODIUM_DIR) && ./autogen.sh && ./configure --disable-shared --enable-static \
+		--disable-debug --disable-dependency-tracking --with-pic --prefix=$(SODIUM_INSTALL)
+	$(MAKE) -C $(LIBSODIUM_DIR)
+	$(MAKE) -C $(LIBSODIUM_DIR) install
```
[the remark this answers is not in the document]
justin1121: > could this interfere with any existing install of libsodium (eg through homebrew)?

No, the configure command above sets the prefix of the installation location to the directory in the user's home path.

Pros:

Easier installation for user and us. We don't need to locate the libsodium installation location whether its install via homebrew/apt or anything else. For people using tfe they don't need to worry about installing extra dependencies.

Cons:

Slower build times.

operations/secure_random/secure_random.cc
```diff
@@ -0,0 +1,203 @@
...
+  explicit SecureRandomOp(OpKernelConstruction* context) : OpKernel(context) {}
+
+  void Compute(OpKernelContext* context) override {
+    const Tensor& shape_t = context->input(0);
+    const Tensor& seed_t = context->input(1);
+    const Tensor& minval = context->input(2);
+    const Tensor& maxval = context->input(3);
+    TensorShape shape;
+    OP_REQUIRES_OK(context, MakeShape(shape_t, &shape));
+    OP_REQUIRES(context, seed_t.dims() == 1 && seed_t.dim_size(0) == 8,
```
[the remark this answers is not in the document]
justin1121: Seed is 32 bytes long so 8 int32s. That's what `randombytes_SEEDBYTES` is, 32 bytes.

operations/secure_random/secure_random.cc
justin1121: Yes. Its 32 bytes. It comes from the libsodium header file.

operations/secure_random/secure_random.cc
justin1121: Good idea. Thing I'm stuck with here is that the output data from `output->flat<T>()` returns a `std::vector` I believe. I couldn't figure out a good way to access the underlying memory allocated for that. We need to do that because libsodium takes a void pointer. We can't just pass a vector through. Any ideas?

operations/secure_random/secure_random.cc
justin1121: Borrowed from libsodium. Here https://github.com/jedisct1/libsodium/blob/master/src/libsodium/randombytes/randombytes.c#L171...

operations/secure_random/secure_random.cc
justin1121: The syntax is `<tensor name>: <tensor type>` and the `<tensor type>` can be references from the attributes below. Thought I'd just call that attribute `dtype` because 3 of the inputs rely on it.

tf_encrypted/operations/secure_random/secure_random.py
justin1121: I was thinking of deferring this sort of thing to other PRs. What do you think?

Makefile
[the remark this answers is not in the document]
mortendahl: perfect, thanks 👍

operations/secure_random/secure_random.cc
```diff
@@ -0,0 +1,203 @@
...
+      context, lo < hi,
+      errors::InvalidArgument("Need minval < maxval, got ", lo, " >= ", hi));
+
+    // Allocate output
+    Tensor* output;
+    OP_REQUIRES_OK(context, context->allocate_output(0, shape, &output));
+    if (shape.num_elements() == 0) return;
+
+    if (sodium_init() < 0) {
+      return;
```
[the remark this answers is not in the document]
mortendahl: could it be something like `OP_REQUIRES(...)`?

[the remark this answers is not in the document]
justin1121: Yep think so.

operations/secure_random/secure_random.cc
mortendahl: my C++ is *very* limited, best I can do is stackoverflow :D https://stackoverflow.com/questions/6485496/how-to-get-stdvector-pointer-to-the-raw-data

tf_encrypted/operations/secure_random/secure_random.py
justin1121: Nope. All inputs come in as tensors as far as I can tell.

operations/secure_random/secure_random.cc
justin1121: This appears to work! Great find!

operations/secure_random/secure_random.cc
justin1121: There's some information on the nonce here: https://download.libsodium.org/doc/advanced/stream_ciphers/chacha20

TL;DR: The stream cipher is created via the ietf version of the chacha20 stream cipher. Its based on a pair of (nonce, key) where the nonce is 96 bits and the key is 256 bits.

operations/secure_random/secure_random.cc
justin1121: rejection sampling was borrowed from here: https://github.com/jedisct1/libsodium/blob/master/src/libsodium/randombytes/randombytes.c#L139

justin1121:
Have some ideas surrounding concurrency! https://github.com/mortendahl/tf-encrypted/issues/270

tf_encrypted/operations/secure_random/secure_random.py
mortendahl: attributes could potentially be used instead, but don't know what best practice is around those

justin1121: Tried doing that but had trouble figuring out how attributes work. The random uniform op in tensorflow passes the min and max values like this as well. Kind of thinking that attributes don't work as inputs at all and just describe an input. Like its type or value constraints.

justin1121: I think what I said is incorrect but there was definitely something weird going on when I added attributes! Can look into it more if you'd like but I am not sure how critical this is!

tests/test_secure_random.py
```diff
@@ -0,0 +1,50 @@
...
+
+            np.testing.assert_array_equal(output, expected)
+
+    def test_min_val(self):
+        expected = [[3220, 200, -8648],
+                    [2223, 9603, 815],
+                    [9356, 4377, 6561]]
+
+        with tf.Session():
+            output = secure_random([3, 3], minval=-10000, maxval=10000).eval()
```
mortendahl: uh, this looks dangerous! :D it will used a fixed seed if none is specified? if so we should definitely change that (potentially requiring a seed to always be specified)!

operations/secure_random/secure_random.cc
```diff
@@ -0,0 +1,247 @@
...
+  uint32 inner_block_index_ = 0;
+
+  // The following random uniform distribution is based on a an implementation from
+  // https://github.com/rust-random/rand/blob/3eadab75c8a5871d1be729091795a6c4e1dc19bb/src/distributions/uniform.rs#L310
+  // There is quite a bit of documentation at that link which is explains the implementation.
+  // See below for other inline docs.
+
+  // inclusive uniform!
+  void Uniform(T low, T high) {
+    typedef typename std::make_unsigned<T>::type uT;
```
mortendahl: this gives the unsigned version of the signed type?

mortendahl: or, how is the fact that the rust code takes four types expressed here?

operations/secure_random/secure_random.cc
mortendahl: any way it could be unified (eg as a `#define`) so we don't have magic numbers (`8`)?

operations/secure_random/test_secure_random.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def test_invalid_max_min(self):
+        with tf.Session():
+            minval = tf.constant(1000, dtype=tf.int64)
+            maxval = tf.constant(-1000, dtype=tf.int64)
+
+            with np.testing.assert_raises(errors.InvalidArgumentError):
+                secure_random_module.secure_random([2, 3], [1, 1, 1, 1, 1, 1, 1, 2], minval, maxval).eval()
+
+    def test_negative_numbers(self):
```
mortendahl: nice! 👍

tf_encrypted/operations/secure_random/secure_random.py
```diff
@@ -0,0 +1,46 @@
...
+    dtype = dtypes.as_dtype(dtype)
+    if dtype not in (dtypes.int32, dtypes.int64):
+        raise ValueError("Invalid dtype %r" % dtype)
+
+    if maxval is None:
+        raise ValueError("Must specify maxval for integer dtype %r" % dtype)
+
+    if seed is None:
+        # TODO random seed if not specified?
+        seed = [87654321, 87654321, 87654321, 87654321, 87654321, 87654321, 87654321, 87654321]
```
mortendahl: ah, I guess this is the line that should be killed per comment elsewhere about not having a default fixed seed

tf_encrypted/operations/secure_random/secure_random.py
```diff
@@ -0,0 +1,46 @@
...
+    elif len(seed) != 8:
```
mortendahl: any way to avoid hardcoding `8` here?

tf_encrypted/operations/secure_random/secure_random.py
mortendahl: matching tf is fine. thanks for looking into it!

tf_encrypted/tensor/crt.py
```diff
@@ -7,6 +7,7 @@
 
 from ..tensor.shared import im2col
 from .helpers import inverse, prod
+from ..operations.secure_random import random_uniform
```
mortendahl: let's stick to tf version for now until we have an unseeded `random_uniform`!

mortendahl: maybe it's just a question of adding a second custom op that doesn't take seed and use that instead; this one could potentially use `/dev/random` or whatever instead to outsource computation to the OS

operations/secure_random/secure_random.cc
justin1121: > this gives the unsigned version of the signed type?

Yes, whether it's int32 or int64.

> or, how is the fact that the rust code takes four types expressed here?

This version doesn't need the four types because in the rust code all four types were being used as an optimizations for smaller width values (int8 and int16). We're only handling int32 and int64 which those optimizations are not needed.

PR #267, opened 2018-10-29
basic skeleton for anaconda package

mortendahl:
cc @ianlivingstone -- did you do any work on this already?

### DIFF
--- a/meta.yaml
+++ b/meta.yaml
@@ -1,0 +1,12 @@
+package:
+  name: tf-encrypted
+  version: "0.2.0"
+
+source:
+  git_url: https://github.com/mortendahl/tf-encrypted.git
+
+  about:
+    home: https://github.com/mortendahl/tf-encrypted
+    license: APACHE-2
+    license_file: LICENSE
+    summary: "Layer on top of TensorFlow for doing machine learning on encrypted data."


ianlivingstone:
Just a small amount of investigation, we'll need to package up the int64 version of Tensorflow into a conda package that we host (conversion from a wheel).

meta.yaml
```diff
@@ -0,0 +1,12 @@
+package:
+  name: tf-encrypted
+  version: "0.2.0"
```
ianlivingstone: This should be added to the Makefile to ensure its updated with the `setup.py` and other files. In future this should probably be templated or something along those lines.

mortendahl: how would you update this using the makefile?

ianlivingstone:
@mortendahl what do we need to do to get this merged up?

mortendahl:
the current version is not taking full advantage for conda but is enough to build a package; merge and come back later?

meta.yaml
```diff
@@ -0,0 +1,18 @@
+package:
+  name: tf-encrypt
```
justin1121: tf-encrypted???

mortendahl: good catch!

mortendahl:
someone wants to accept this so the PR can be closed? @justin1121?

mortendahl:
@justin1121 done

PR #280, opened 2018-11-07
platform specific

justin1121:


.circleci/config.yml
```diff
@@ -18,7 +18,7 @@ variables:
   build-release: &build-release
     run:
       name: Build and push release artifacts (e.g. docker containers)
-      command: RELEASE_CONFIRM=yes make push
+      command: RELEASE_CONFIRM=yes PYPI_PLATFORM=manylinux1_x86_64 make push
```
justin1121: While we don't support manylinux, as far as I can pypi will only let you upload packages tagged with manylinux. Tensorflow is our main dependency and they also tag their releases with manylinux but don't actually support manylinux. This is why we're held back from actually supporting manylinux. If a user can find a tensorflow distribution 1.9 or greater that works on their machine they should be able to use tfe.

.circleci/config.yml
```diff
@@ -64,6 +64,25 @@ jobs:
...
+          command: |
+            curl -O https://repo.continuum.io/miniconda/Miniconda3-latest-MacOSX-x86_64.sh
+            bash Miniconda3-latest-MacOSX-x86_64.sh -b
+            ../miniconda3/bin/conda create -n py35 python=3.5
+            ../miniconda3/envs/py35/bin/python -m venv venv
+            . venv/bin/activate
+            make bootstrap
+      - run:
+          name: release
+          command: RELEASE_CONFIRM=yes PYPI_PLATFORM=macosx_10_11_x86_64 make pypi-push
```
justin1121: no docker builds for macos

.circleci/config.yml
```diff
@@ -64,6 +64,25 @@ jobs:
...
+  deploy-macos:
+    macos:
+        xcode: "10.0.0"
+    working_directory: ~/repo
+    steps:
+      - checkout
+      - run:
+          name: bootstrap setup
+          command: |
+            curl -O https://repo.continuum.io/miniconda/Miniconda3-latest-MacOSX-x86_64.sh
```
ianlivingstone: Does Circle CI have os x machines with conda already installed? Perhaps we could cache these dependencies?

.circleci/config.yml
```diff
@@ -75,6 +94,14 @@ workflows:
               # the build job on all tags. By default, Circle CI only kicks off
               # builds on tags if a filter is defined.
               only: /^(?:[0-9]+)\.(?:[0-9]+)\.(?:[0-9]+)(?:(\-rc[0-9]+)?)$/
+      - deploy-macos:
```
ianlivingstone: Is there a way for us to only release if the other artefacts have been generated properly?

Makefile
```diff
@@ -218,10 +219,15 @@ ifeq (,$(shell grep -e $(VERSION) docs/source/conf.py))
 	$(error "Version specified in docs/source/conf.py does not match $(VERSION)")
 endif
 
-pypi-push-master: pypicheck pypi-version-check
+pypi-platform-check:
+ifeq (,$(PYPI_PLATFORM))
```
ianlivingstone: maybe this should just have a default if its not provided?

.circleci/config.yml
justin1121: Don't think so. Will look into caching.

.circleci/config.yml
justin1121: Can we do:

```
requires:
  - build
  - deploy
```
??

Makefile
justin1121: sure

.circleci/config.yml
ianlivingstone: We can have it be a three part pipeline I think:

- build
- create-artifacts across all 3 platforms (windows, mac os x, etc)
- publish

Just not sure how to pass assets from creating the artifacts to publish.

.circleci/config.yml
ianlivingstone: How'd we do this before now? I don't think we passed manylinux at all previously!

justin1121: We weren't explicitly platform specific before. But we were implicitly restricted to whatever tensorflow supports.

.circleci/config.yml
justin1121: fyi circleci doesn't have windows machines

.circleci/config.yml
justin1121: O I think I understand your question now! I should be more specific. PYPI doesn't allow linux builds e.g. linux_x86_64 or something like ubuntu_x86_64. It only allows manylinux1_x86_64 for linux builds. We were just passing any platform before which was fine.

ianlivingstone: Ah, got you! Is this documented someplace? Perhaps in the Makefile?

.circleci/config.yml
ianlivingstone: Can you open a new issue for us to tackle this in future?

justin1121: Also to do with your original point. I can't really find any way of storing artifacts. While it's a nice to have I don't know if it's critical right now. I can open an issue about it to look into it in the future.

PR #281, opened 2018-11-07
Fix for private compare bug

mortendahl:
Thanks for spotting this @morgangiraud 🙏


morgangiraud:
Congrats on fixing it 💯 

Happy to see that the investigation was worthed it 👍

tests/morgan-goodbye-present.py
```diff
@@ -0,0 +1,40 @@
+import numpy as np
```
justin1121: Should we turn this file into a real test?!

tests/test_int100_tensor.py
```diff
@@ -32,34 +34,56 @@ def test_pond(self) -> None:
             expected = np.array([4, 4])
             np.testing.assert_array_almost_equal(actual, expected, decimal=3)
 
-    def test_binarize(self) -> None:
+    def core_test_binarize(self, raw, shape, modulus, bitlen, ensure_positive_interpretation) -> None:
```
justin1121: Wonder if this needs to start with test to actually run as a test.

tests/test_truncate.py
```diff
@@ -18,15 +18,17 @@ def test_interactive_truncate(self):
             fixedpoint_config=fixed100,
         )
 
-        with tfe.Session() as sess:
+        # TODO[Morten] remove this condition
+        if prot.tensor_factory not in [tfe.tensor.int64.int64factory]:
```
justin1121: I assume this is because interactive truncation doesn't work with int64. Maybe we should open an issue?

tf_encrypted/protocol/pond.py
```diff
@@ -1554,6 +1555,18 @@ def _type(x):
     return type(x)
 
 
+# TODO[Morten] this is just a very first step; far from finished
+def debug(x: PondTensor, summarize=None, message=''):
```
justin1121: Awesome!!

tests/morgan-goodbye-present.py
mortendahl: good catch! i think i'll delete it, the error was captured by relu tests anyway

PR #289, opened 2018-11-12
Resolved "Add public division to POND #126"

Nivek92:
Added div functionality to Pond, PondTensor and various backing tensor types.

I more or less just copied the mul functionality and replaced the multiplication with division.

### DIFF
--- a/tf_encrypted/protocol/pond.py
+++ b/tf_encrypted/protocol/pond.py
@@ -1109,6 +1109,22 @@ class PondTensor(abc.ABC):
     def __rmul__(self, other):
         return self.prot.mul(self, other)
 
+    def div(self, other):
+        """
+        Divide this tensor with `other`
+
+        :param PondTensor other: to divide
+        :return: A new PondTensor
+        :rtype: PondTensor
+        """
+        return self.prot.div(self, other)
+
+    def __div__(self, other):
+        return self.prot.div(self, other)
+
+    def __rdiv__(self, other):
+        return self.prot.div(self, other)
+
     def __mod__(self, other):
         return self.prot.mod(self, other)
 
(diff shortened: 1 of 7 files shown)


Nivek92:
@mortendahl ok good to know. will look into it and fix it accordingly

mortendahl:
I think the easiest is to only support public division with an int or floating point for now, ie you can do `x / y` when:

- `y` is an integer
- `y` is a float
- `y` is np.ndarray

but otherwise *not* (ie leave out the case where `y` is `PondPublicTensor` or `PondPrivateTensor`).

That way there is no need to decode first, you can simply compute the reciprocal and use the existing methods for multiplication. Once working it's not too much trouble to add support for `PondPublicTensor`.

Very happy to see the interest! 👍 🎉

Nivek92:
So heres my understanding.

In Int64Tensor & Int32Tensor I can use 

```
    def div(self, other: Any) -> 'IntXTensor':
        x, y = _lift(self, other)
        return intXfactory.tensor(x.value / y.value) 
```

as x.value and x.value are either int or tf.Tensor

for Int100Tensor which is the representation in a residue number system
you can multiply with the multiplicative inverse but only if Y and M are coprimes
otherwise you have to use some fancy algorithms thus I would use raise a NotImplementedError in that case

for PrimeTensor as the modular (following known as p) is a prime we can use y = pow(x, p-2, p) to get the multiplicative inverse for y and then just do  x * y % p

and then all the implementations of _div_public_public etc. should work as they delegate to the protocol which delegates down to the backing tensors correct?

mortendahl:
The multiplicative ring inverse is not exactly what we're after here, it's rather the fixed point inverse!

mortendahl:
@Nivek92 want to add a few tests?

PR #291, opened 2018-11-19
Bump flake8 and fix linting

jvmancuso:
Updates:
- flake8, 3.5.0 --> 3.6.0
- Fix mac build caching in CircleCI
- Increment dependency cache version
- Ignore W504 project-wide in favor of enforcing W503 (when necessary, line breaks go before binary ops instead of after)
- Fix some linter errors

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,25 +22,26 @@ variables:
   restore-cache: &restore-cache
     restore_cache:
       keys:
-        - v1-dependencies-{{ checksum "requirements.txt" }}
+        - v2-dependencies-{{ checksum "requirements.txt" }}
         # fallback to using the latest cache if no exact match is found
-        - v1-dependencies-
+        - v2-dependencies-
   save-cache: &save-cache
     save_cache:
       paths:
         - ./venv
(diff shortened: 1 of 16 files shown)


.circleci/config.yml
```diff
@@ -22,25 +22,26 @@ variables:
   restore-cache: &restore-cache
     restore_cache:
       keys:
-        - v1-dependencies-{{ checksum "requirements.txt" }}
+        - v2-dependencies-{{ checksum "requirements.txt" }}
```
ianlivingstone: Why do we need to increment the deps cache?

jvmancuso: seemed an easy way to force a cache rebuild, but then realized it was because the mac dependencies cache didn't have a fallback for auto-rebuilding -- can revert if we prefer?

ianlivingstone: Refer would be preferable since this should invalidate the cache when the requirements file changes!

jvmancuso: we decided to keep the increment, since circleci fails without it

justin1121: yeah there's a messed up cache in there so this is the best solution unless we contact circleci and tell them to purge it.

Issue #200, opened 2018-09-29
Implement 1-D Convolutions


ianlivingstone:
We'd like to implement 1-D Convolutions and then use them to evaluate and benchmark our support for sequence-based models (e.g. text classification with bag of 1d-convolutions, etc.).

Depends on #358

ianlivingstone:
@jvmancuso could you add some more colour to this issue? :)

jvmancuso:
## Plan
- [ ] Implement missing ops with necessary conversion logic (below)
- [ ] Audit and tweak Conv2d, as needed (we're not 100% confident that the original conv2d layer was implemented to handle non-standard input tensors)
- [ ] Debug and test character-level language model conversion into TFE

### Missing Ops:
- [ ] [`tf.pad`](https://www.tensorflow.org/api_docs/python/tf/pad)
- [ ] [`tf.required_space_to_batch_paddings`](https://www.tensorflow.org/api_docs/python/tf/required_space_to_batch_paddings)
- [ ] [`tf.stack`](https://www.tensorflow.org/api_docs/python/tf/stack) (previously known as `tf.pack`, still shows up in protobuf as a `Pack` op
- [ ] [`tf.concat`](https://www.tensorflow.org/api_docs/python/tf/concat)
- [ ] [`tf.batch_to_space_nd`](https://www.tensorflow.org/api_docs/python/tf/batch_to_space_nd)
- [ ] [`tf.space_to_batch_nd`](https://www.tensorflow.org/api_docs/python/tf/space_to_batch_nd)

kamathhrishi:
Could you elaborate more on what non standard input tensors could be like ?

jvmancuso:
Standard for Conv2D would likely be something like an input of shape (n, d, h, w) where usually h=w, but with Conv1D, the inputs to Conv2d are reshaped to (n, d, h, 1) or something similar. We'll just want to make sure our current Conv2d can support this shape of input.

Once #297 is merged we'll be able to convert a Conv1D layer into tf-encrypted, but if our Conv2D implementation can't handle that input shape, we'll probably run into a few errors when we try.  Note we'll likely have a PR for #297 in progress by end of day or early tomorrow

PR #299, opened 2018-11-21
Pad

yanndupis:
Related #200 

Hey @jvmancuso, @justin1121,

This PR is adding the [`tf.pad`](https://www.tensorflow.org/api_docs/python/tf/pad) operation. Currently it just pads tensors with 0s (private) however it should be sufficient for our conv1d use case.

Thank you!


tf_encrypted/protocol/pond.py
```diff
@@ -989,6 +989,71 @@ def dispatch(self, base_name, *args, container=None, **kwargs):
         else:
             raise TypeError("Don't know how to {}: {}".format(base_name, [type(arg) for arg in args]))
 
+    @memoize
+    def zeros(
```
justin1121: I think we'll want to implement the zeros functions for public, private and masked tensors. Not sure if we need to for pad though might be able to infer the correct type based on whatever is passed in and then used in zeros.

yanndupis: Hey Justin, good point. I have added support for public and masked for zeros. Let me know if that's what you had in mind. Thank you!

tf_encrypted/protocol/pond.py
```diff
@@ -989,6 +989,142 @@ def dispatch(self, base_name, *args, container=None, **kwargs):
         else:
             raise TypeError("Don't know how to {}: {}".format(base_name, [type(arg) for arg in args]))
 
+    @memoize
+    def zeros_private(
```
jvmancuso: thinking maybe we want to combine these into one function?  I know the normal method of dispatching won't work the same way here, but the code is so similar that it feels like it'd be better to pass a kwarg in that determines if the output is public, private, or masked and have it return the corresponding tensor type

tests/test_convert.py
```diff
@@ -821,6 +857,25 @@ def export_expand_dims(filename: str, input_shape: List[int]):
     return export(x, filename)
 
 
+def run_pad(input_shape: List[int]):
```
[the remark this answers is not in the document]
yanndupis: good catch

tests/test_pad.py
```diff
@@ -0,0 +1,43 @@
...
+import numpy as np
+import tensorflow as tf
+import tf_encrypted as tfe
+
+
+class TestPad(unittest.TestCase):
+    def setUp(self):
+        tf.reset_default_graph()
+
+    def test_zeros(self):
```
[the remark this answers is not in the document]
yanndupis: good catch!

tf_encrypted/protocol/pond.py
```diff
@@ -989,6 +989,77 @@ def dispatch(self, base_name, *args, container=None, **kwargs):
...
+        self,
+        shape,
+        tensor_type: str = "private",
+        apply_scaling: bool = True,
+        name: Optional[str] = None,
+        factory: Optional[AbstractFactory] = None
+    ) -> Union['PondPublicTensor', 'PondPrivateTensor', 'PondMaskedTensor']:
+
+        if tensor_type == "public":
+            out = _zeros_public(self, shape, apply_scaling, name, factory)
```
justin1121: Hey Yann, you can just check the tensor type here like: https://github.com/mortendahl/tf-encrypted/blob/master/tf_encrypted/protocol/pond.py#L869

tf_encrypted/protocol/pond.py
```diff
@@ -989,6 +989,77 @@ def dispatch(self, base_name, *args, container=None, **kwargs):
...
+
+    def _prepend_zeros(self, arr: np.ndarray, pad_amt: int, axis: int, zeros_id: int):
+
+        if pad_amt == 0:
+                return arr
+
+        arrshape = arr.shape.as_list()
+        padshape = tuple(x if i != axis else pad_amt for (i, x) in enumerate(arrshape))
+
+        if isinstance(arr, PondPublicTensor):
```
justin1121: related to above, you shouldn't need to check the instances and pass a specific tensor type, it should just pick the right one automatically if you have the base zeros set up correctly

tests/test_convert.py
```diff
@@ -59,7 +59,7 @@ def provide_input():
             x = converter.convert(graph_def, register(), 'input-provider', provide_input)
 
             with tfe.Session() as sess:
-                sess.run(prot.initializer, tag='init')
+                sess.run(tf.global_variables_initializer())
```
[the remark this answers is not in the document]
yanndupis: I changed it to make it consistent with the other tests. Also when zeros tensors are created, under the hood it's creating a variable. Without the modification above, I was getting the error `Attempting to use uninitialized value pad/private-zeros-1/Variable_1`. I could maybe create a PondConstant instead, however we are currently just supporting public PondConstant. That's something I could change in a separate PR if you think that's more appropriate. Thanks,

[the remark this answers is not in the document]
justin1121: Feel like consistency is good! Also the tag kwarg only comes into play when examining the graphs with tensorboard.

tf_encrypted/protocol/pond.py
yanndupis: Hey Justin, how would you do that? zeros doesn't take a tensor as an input. It just takes a shape as an input then create a tensor with zeros based on this shape. So would you still add a tensor as an input like `def zeros(self, shape, input: PondTensor,` thanks for your review.

justin1121: Oh sorry didn't notice this didn't take any input. Jumped the gun!

justin1121: Wonder if there is another way to pass in a type instead of a string though.

yanndupis: No worries, thanks for your question. good point, let me check

jvmancuso: Hey, wondering if we actually need an explicit masked version of this function or if public and private would be sufficient...feel like if we needed it masked for any reason it would be lifted at that point

tests/test_convert.py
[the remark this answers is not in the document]
jvmancuso: Yup consistency is what I was pushing for, but makes sense to remove the tag everywhere now that you mention the tensorboard thing!

tf_encrypted/protocol/pond.py
yanndupis: That's a good point. Let me remove mask.

yanndupis:
Hey guys, I think this pr should be ready to be merged. Thank you.

Issue #300, opened 2018-11-21
deploy-macos on circleci fails because wget not installed


justin1121:
Just use curl!!!!

```
wget -nc https://github.com/jedisct1/libsodium/archive/1.0.16.tar.gz
/bin/bash: wget: command not found
```

PR #302, opened 2018-11-22
use curl instead of wget

justin1121:
Fixes https://github.com/mortendahl/tf-encrypted/issues/300

We might want to start doing macos builds (not deploys) on every commit to avoid these sort of bugs. I just worry about builds for each commit taking too long and using up all our macos build machines (I think we only have 1 right now).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -284,7 +284,7 @@ push: pypi-version-check
 # ###############################################
 
 $(LIBSODIUM_OUT):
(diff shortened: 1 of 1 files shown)


ianlivingstone:
> We might want to start doing macos builds (not deploys)

Building on every commit in a PR makes sense for us now, we can always pay to upgrade our parallelism. What's most important is that we catch bugs early and do our best to prevent them from being introduced into master :)

justin1121:
Sure, I agree. I'll set that up now just only take a second!

PR #304, opened 2018-11-23
Custom ops tf version

justin1121:
This builds different shared objects for each version of tensorflow we support and then packages them up into the pip package when we deploy.


justin1121:
Done, thanks! https://github.com/mortendahl/tf-encrypted/issues/307

Issue #202, opened 2018-10-01
SecureNN with any backing tensor


justin1121:
When the initial SecureNN MVP as described here (https://github.com/mortendahl/tf-encrypted/issues/106) is completed we want to expand on this and make it so it can work with any backing tensor (e.g. `Int32Tensor` and `Int64Tensor`). The initial MVP will be finished with only CRT support. 

### Next Steps

- [x] share convert (should be completed as part as MVP)
- [x] add proper support for encoding numbers to fixed point representations inside the int32 and int64 tensors
- [ ] test secure with int32 and int64 tensors

mortendahl:
skipping int32 support for now (although nothing should prevent it from working)

Issue #1, opened 2018-06-05
64bit tensors with natural mod reductions


mortendahl:
For some applications 64bit might be enough, in which case we can not only avoid the use of the CRT also potentially switch to "native modulo reductions" in the form of natural wrap-arounds as opposed to explicit `%` operations. This combined setting is for instance used in [SecureML](https://eprint.iacr.org/2017/396).

jvmancuso:
More discussion in #100

PR #306, opened 2018-11-23
Adds int64 support for SecureNN

mortendahl:
closes https://github.com/mortendahl/tf-encrypted/issues/1
closes https://github.com/mortendahl/tf-encrypted/issues/202

### DIFF
--- a/tf_encrypted/protocol/pond.py
+++ b/tf_encrypted/protocol/pond.py
@@ -37,13 +37,15 @@ _thismodule = sys.modules[__name__]
 
 class Pond(Protocol):
     """
+    Pond(server_0, server_1, crypto_producer, tensor_factory, fixedpoint_config)
+
     Pond is similar to SPDZ except it has been vectorized plus a few more optimizations.
 
     Pond works with 2 parties for computation and one crypto producer for triples.
 
     :param Player server_0: The "alice" of MPC.
     :param Player server_1: The "bob" of MPC.
-    :param Player crypto_producer: The host to act as the crypto producer.  In pond this party is
+    :param Player crypto_producer: The host to act as the crypto producer.  In `Pond` this party is
         responsible for producing triples to aid in computation.
     :param AbstractFactory tensor_factory: Which backing type of tensor you would like to use, e.g. `int100` or `int64`
     """  # noqa:E501
@@ -93,6 +95,8 @@ class Pond(Protocol):
         factory: Optional[AbstractFactory] = None,
     ) -> "PondConstant":
         """
+        define_constant(value, apply_scaling, name, factory) -> PondConstant
+
         Define a constant to use in computation.
 
         .. code-block:: python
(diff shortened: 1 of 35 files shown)


tf_encrypted/protocol/pond.py
```diff
@@ -3512,3 +3510,22 @@ def _cast_backing_public(
             y_on_1 = x_on_1.cast(backing_dtype)
 
         return PondPublicTensor(prot, y_on_0, y_on_1, x.is_scaled)
+
+
+def _cast_backing_private(prot: Pond, x: PondPrivateTensor, backing_dtype) -> PondPrivateTensor:
```
justin1121: Wonder if this should be moved to securenn if its not being used elsewhere.

tf_encrypted/protocol/securenn/securenn.py
```diff
@@ -458,50 +443,67 @@ def _bits_public(prot, x: PondPublicTensor,
...
 
     with tf.name_scope('lsb'):
 
         with tf.name_scope('lsb_mask'):
 
             with tf.device(prot.server_2.device_name):
-                x_raw = y.backing_dtype.sample_uniform(y.shape)
-                xbits_raw = x_raw.to_bits(factory=prot.prime_factory)
-                xlsb_raw = xbits_raw[..., 0]
+                r0_raw = odd_dtype.sample_uniform(x.shape)
```
justin1121: Curious why not just `_share`?

tf_encrypted/tensor/int100.py
```diff
@@ -379,6 +380,10 @@ def truncate(self, amount, base=2):
     def right_shift(self, bitlength):
         return self.truncate(bitlength, 2)
 
+    def cast(self, factory):
```
justin1121: This is saying int100tensors don't need to be casted so just return ourself?

mortendahl: the assertion makes sure that you can only cast int100 it to int100; it's a bit ugly and should be cleanup up once we have more advanced types for tensors!

mortendahl: for now it's needed to use the same code for securenn on int64 and int100 (where the former is cast to the odd tensor)

tf_encrypted/protocol/securenn/securenn.py
mortendahl: good catch, believe we could change it now if that's better.

motivation was that `_share` works differently: it draws one sample and subtracts that from the value. here we are essentially sharing a random value, so doing it this way gives exactly the same; however, no subtraction is involved. in the end I implemented subtraction for odd tensors as well so `_share` could be used, yet this is likely how we'll share random tensors after introducing seeding.

tf_encrypted/protocol/pond.py
mortendahl: yeah I like that, not least since it's half-baked for now. will move both public and private casting

mortendahl: done

PR #314, opened 2018-11-27
BatchToSpaceND and SpaceToBatchND

jvmancuso:
Related to #200

Implements the `tf.batch_to_space_nd` and `tf.space_to_batch_nd` ops at various levels of abstraction, including Pond, Int100Tensor, Int64Tensor, Int32Tensor, PrimeTensor.

Other stuff:
- While rebuilding tfe from the recent secure random work, found a few bugs for `make boostrap` on my system.
    - We were forcing a specific version of `certifi` that was outdated with the standard version installed by conda for use with python 3.5.
    - We were also forcing tensorflow 1.9.0, even though we'd last planned on allowing >= 1.9.0 according to https://github.com/mortendahl/tf-encrypted/pull/236#issue-222671228.
- Added documentation for the `player/` subdirectory that I forgot to include in a PR last week.


tf_encrypted/player/player.py
```diff
@@ -4,6 +4,16 @@
...
 
 class Player(object):
+    """
+    Player(name, index, device_name, host=None)
+
+    An abstraction for players in the game-theoretic model of a secure computation.
+
+    :param str name: Name of the player
+    :param int index: Index of the player (for ordering)
+    :param str device_name: Name of device (e.g. server_0)
```
mortendahl: `device_name` is the fully expanded name, not `server_0`

tf_encrypted/protocol/pond.py
```diff
@@ -2781,6 +2787,85 @@ def _avgpool2d_masked(
         return PondPrivateTensor(prot, y_on_0, y_on_1, x.is_scaled) * scalar
 
 
+#
+# batch_to_space_nd and space_to_batch_nd helpers
+#
+
+
+def _batch_to_space_nd_core(prot, tensor, block_shape, crops):
+    x_on_0, x_on_1 = tensor.unwrapped
```
mortendahl: maybe rename `tensor` to `x` to match with `x_on_0` etc

mortendahl: applies to `_space_to_batch_nd_core` as well

tf_encrypted/protocol/securenn/odd_tensor.py
```diff
@@ -160,6 +160,12 @@ def im2col(self, h_filter, w_filter, padding, strides) -> 'OddImplicitTensor':
     def conv2d(self, other, strides, padding='SAME') -> 'OddImplicitTensor':
         raise NotImplementedError()
 
+    def batch_to_space_nd(self, block_shape, crops):
```
mortendahl: could we leave these out? is something complaining if we don't have them? trying to avoid this tensor type getting bloated

tf_encrypted/protocol/pond.py
jvmancuso: would probably rather go the other way -- `tensor` is more informative than `x`

mortendahl: jup, either works!

tf_encrypted/protocol/securenn/odd_tensor.py
mortendahl: I see that the odd tensor as being hidden and not for use elsewhere, so think it's okay to get hard errors :)

PR #315, opened 2018-11-27
release 0.4.0 rc0

justin1121:


CHANGELOG.md
```diff
@@ -6,6 +6,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.4.0]
+
+**Added**
+
+- SecureNN with int64 support has landed.
```
mortendahl: maybe mark somewhere that a nightly/custom version of TF is needed for int64 (but CRT works in either case)

CHANGELOG.md
```diff
@@ -6,6 +6,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Cryptographically secure random numbers feature has been implemented but not integrated.
```
mortendahl: maybe it's worth switching to the un-seeded secure random ops before releasing? would it require a lot? then we could avoid the "but not integrated" part

CHANGELOG.md
```diff
@@ -6,6 +6,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Three new ops have been added to Pond and the converter: Pad, BatchToSpaceND, SpaceToBatchND
+
+**Changed**
+
+- There are now separate wheels published to pypi for MacOS and linux.
```
ianlivingstone:
```suggestion
- There are now separate wheels published to PyPI for MacOS and Linux.
```

CHANGELOG.md
```diff
@@ -6,6 +6,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.4.0]
+
+**Added**
+
+- SecureNN with int64 support has landed.
+- Cryptographically secure random numbers feature has been implemented but not integrated.
+- Three new ops have been added to Pond and the converter: Pad, BatchToSpaceND, SpaceToBatchND
```
ianlivingstone: Perhaps just mention we now support three new operations?

CHANGELOG.md
justin1121: O yeah. Thanks.

CHANGELOG.md
justin1121: Sure lets do it. And then we can remove the part from the README too.

PR #325, opened 2018-11-28
fix rejection sampling, call the correct subclass method

justin1121:


Makefile
```diff
@@ -281,6 +281,7 @@ SODIUM_INSTALL = $(shell pwd)/build
 SECURE_OUT_PRE = $(PACKAGE_DIR)/secure_random/secure_random_module_tf_
 
 SECURE_IN = operations/secure_random/secure_random.cc
+SECURE_IN_H = operations/secure_random/generators.h
```
mortendahl: I could see this naming running loose; is there a way to not have to specify all custom ops? Or at least use more telling names? :P

justin1121: Oh yeah there is definitely some Makefile magic to not have to specify every custom op like this. We can also make use of bazel to build custom ops. It might be interesting to look into this once we have more custom ops to manage!!

PR #326, opened 2018-11-28
Floormod2

yanndupis:
Hey guys, 

Here is a public implementation of floormod. 

Thank you!


tf_encrypted/convert/register.py
```diff
@@ -385,6 +387,40 @@ def space_to_batch_nd(converter, node, inputs):
...
+
+    warnings.warn(WARNING, UserWarning)
+
+    x = converter.outputs[inputs[0]]
+    y = converter.outputs[inputs[1]]
+
+    if isinstance(x, tf.NodeDef):
+        x_out = nodef_to_public_pond(converter, x)
+    else:
+        if isinstance(x, PondPrivateTensor):
```
justin1121: Is this private case ever actually happening?

yanndupis: Hey Justin, thank you for your review. in fact yes. if x in the graph is a placeholder, `x = converter.outputs[inputs[0]]` will generate a PrivateTensor. For y, I don't think it could be a private tensor because it seems to be always an attribute on the graph. Should I remove this condition for y?

However one thing I realized I need to change is `out = converter.protocol.define_public_input(converter.model_provider, inputter_fn)` to `converter.protocol.define_private_input`. The output of this operation should be private. Also if the results get stacks with other tensors and if there is a mix of private and public tensors, it will create some issues in the stack ops. Thanks

justin1121: Hmmm, I feel like we should be restricting the input into this floormod implementation to only be public. If the intention is to only be working on shapes at this point for "required space to batch" then those should always be public. Same with the output we shouldn't be outputting private shapes..

yanndupis: hmmm sounds good for the input. For the output, I am not 100%. Would you make then all the tensors public within `required space to batch`? because I don't think we can mix public and private tensor for `pack/stack`. Also if we want to make everything public within `required space to batch`, it's going to be a challenge because it's using lot of common ops and not sure how we could indicate do this ops publicly if it comes from `required space to batch`. I am probably missing something. Here a the ops generated by `required space to batch`.

```
Name: layer1_conv2/required_space_to_batch_paddings/base_paddings ops: Const
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice/stack ops: Const
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice/stack_1 ops: Const
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice/stack_2 ops: Const
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice ops: StridedSlice
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice_1/stack ops: Const
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice_1/stack_1 ops: Const
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice_1/stack_2 ops: Const
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice_1 ops: StridedSlice
Name: layer1_conv2/required_space_to_batch_paddings/add ops: Add
Name: layer1_conv2/required_space_to_batch_paddings/add_1 ops: Add
Name: layer1_conv2/required_space_to_batch_paddings/mod ops: FloorMod
Name: layer1_conv2/required_space_to_batch_paddings/sub ops: Sub
Name: layer1_conv2/required_space_to_batch_paddings/mod_1 ops: FloorMod
Name: layer1_conv2/required_space_to_batch_paddings/add_2 ops: Add
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice_2/stack ops: Const
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice_2/stack_1 ops: Const
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice_2/stack_2 ops: Const
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice_2 ops: StridedSlice
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice_3/stack ops: Const
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice_3/stack_1 ops: Const
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice_3/stack_2 ops: Const
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice_3 ops: StridedSlice
Name: layer1_conv2/required_space_to_batch_paddings/paddings/0 ops: Pack
Name: layer1_conv2/required_space_to_batch_paddings/paddings ops: Pack
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice_4/stack ops: Const
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice_4/stack_1 ops: Const
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice_4/stack_2 ops: Const
Name: layer1_conv2/required_space_to_batch_paddings/strided_slice_4 ops: StridedSlice
Name: layer1_conv2/required_space_to_batch_paddings/crops/0/0 ops: Const
Name: layer1_conv2/required_space_to_batch_paddings/crops/0 ops: Pack
Name: layer1_conv2/required_space_to_batch_paddings/crops ops: Pack
```

jvmancuso: just wanted to mention that the Pack ops will never happen on private tensors -- all of these tensors are int32 tensors that represent shapes.  the output of these are the `crops` and `paddings` args to `batch_to_space_nd`.  here's the line in tf source code for `required_space_to_batch_paddings` -- https://github.com/tensorflow/tensorflow/blob/a6d8ffae097d0132989ae4688d224121ec6d8f35/tensorflow/python/ops/array_ops.py#L2242

jvmancuso: point being, these specific Pack ops will only ever run on lists of public tensors

yanndupis:
Following up our discussion, `required_space_to_batch_paddings` it's generating lots of sub ops and it's going to be challenging to deal with this complexity because of int32 etc. So we have decided to implement `required_space_to_batch_paddings` at higher level. So we don't need to implement floormod anymore. 

Thank you!

Issue #137, opened 2018-09-19
Use seeded randomness


mortendahl:
Reduce networking by sending seeds instead of expanded tensors when possible.

WIP: https://github.com/mortendahl/tf-encrypted/tree/seeds
Issue posted to tf: https://github.com/tensorflow/tensorflow/issues/21974

mortendahl:
blocked waiting for either feedback from tensorflow team or getting custom ops running

justin1121:
Hey @mortendahl. I discovered this set of random functions that take seeds as tensors. https://www.tensorflow.org/api_docs/python/tf/contrib/stateless/stateless_random_uniform

justin1121:
Investigated this a little bit and I'm not quite it will be useful because it can't return ints only floats. We would have to scale the floats to the maximum range and according to @mortendahl this will not provide an even distribution over all the available ints.

One way forward would be to look at the ops for returning random ints and this new stateless random op and borrow some code from each to make our own custom op that does exactly what we need.

Random Uniform Int Op: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/random_op.cc#L217

You can see how it creates a distribution for low to high here: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/random_op.cc#L250

StateLess Random Base Op: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/stateless_random_ops.cc#L67

Implementation of Fill for the base Op: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/stateless_random_ops.cc#L105

And you can see here the `Distribution` not being created a low or high value: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/stateless_random_ops.cc#L110

mortendahl:
think this is exactly the approach we should take (using a CPRG)

ianlivingstone:
Assigning @justin1121 and moving out of blocked.

We want to spend some time researching custom ops and how we'd package them into our distribution (the artifact we push up to pypi) and how it'd affect the installation experience for our users. Most importantly, how would this impact our own build pipeline? Would we need to do something extra to build and package up the custom op?

Furthermore, what's block our ability to implement a custom op for random seeds today? Anything? Are there other speed ups or use cases for custom ops that'd enable us to realize performance improvements?

Once we have a good understanding of the above questions we can start implementation!

ianlivingstone:
[...]

PR #327, opened 2018-11-29
Seeded tensors

justin1121:
Closes https://github.com/mortendahl/tf-encrypted/issues/137


tests/test_secure_random.py
```diff
@@ -77,13 +84,15 @@ def test_invalid_args(self):
...
         s = seed_gen()
+        s0 = seed_gen()
 
         minval = -2000
         maxval = 0
 
         shape = [2, 3]
 
         output = seeded_random_uniform(shape, seed=s, minval=minval, maxval=maxval).eval()
+        output2 = seeded_random_uniform(shape, seed=s0, minval=minval, maxval=maxval).eval()
```
mortendahl: is `output2` ever used?

tf_encrypted/protocol/pond.py
```diff
@@ -719,6 +730,22 @@ def mask(self, x):
...
         return x_masked
 
+    def expand(self, x):
+        node_key = ("expand", x)
+        x_expanded = nodes.get(node_key, None)
+
+        if x_expanded is not None:
+            return x_expanded
+
+        if isinstance(x, PondMaskedTensor):
```
mortendahl: something is missing here right?

tf_encrypted/protocol/pond.py
```diff
@@ -3377,31 +3432,46 @@ def _mask_private(prot: Pond, x: PondPrivateTensor) -> PondMaskedTensor:
...
+            s0 = seed()
+            s1 = seed()
+            a0seed = x.backing_dtype.seeded_tensor(x.shape, s0)
+            a1seed = x.backing_dtype.seeded_tensor(x.shape, s1)
+
+        return PondMaskedTensor(prot, x, None, a0seed, a1seed, None, None, x.is_scaled)
+
+
+def _expand_masked(prot: Pond, x: PondMaskedTensor) -> PondMaskedTensor:
+    if x.expanded:
```
mortendahl: rename to `is_expanded` to match `is_scaled`?

tf_encrypted/protocol/pond.py
```diff
@@ -573,6 +572,19 @@ def _share_and_wrap(self, secret: AbstractTensor, is_scaled) -> "PondPrivateTens
...
 
+    def _share_one_seed(self, secret):
+        with tf.name_scope("share_one_seed"):
+            s = seed()
+            share0 = secret.factory.seeded_tensor(secret.shape, s)
+            share1 = secret - share0
+
+            # randomize which party gets the seed
+            r = random()
+            if r > 0.5:
```
mortendahl: there's a bias here, should be "random() < 0.5" since the value is in `[0, 1)`

mortendahl: (not important for security though)

tf_encrypted/protocol/pond.py
```diff
@@ -3258,6 +3309,8 @@ def _stack_masked(
 ) -> PondMaskedTensor:
     assert all(x.is_scaled for x in xs) or all(not x.is_scaled for x in xs)
 
+    xs = [prot.expand(x_i) for x_i in xs]
```
mortendahl: don't think we have to expand here, ie we could wait

tf_encrypted/protocol/pond.py
```diff
@@ -3337,6 +3390,8 @@ def _concat_masked(
...
```
mortendahl: same here, not sure we need to expand yet

tf_encrypted/protocol/pond.py
```diff
@@ -3377,31 +3432,46 @@ def _mask_private(prot: Pond, x: PondPrivateTensor) -> PondMaskedTensor:
     x0, x1 = x.unwrapped
 
     with tf.name_scope("mask"):
+        with tf.device(prot.crypto_producer.device_name):
+            s0 = seed()
```
mortendahl: could be inlined as it's not needed on it's own (leaving eg only `a0` and `a1`)

tf_encrypted/protocol/pond.py
```diff
@@ -3377,31 +3432,46 @@ def _mask_private(prot: Pond, x: PondPrivateTensor) -> PondMaskedTensor:
...
+            s1 = seed()
+            a0seed = x.backing_dtype.seeded_tensor(x.shape, s0)
+            a1seed = x.backing_dtype.seeded_tensor(x.shape, s1)
+
+        return PondMaskedTensor(prot, x, None, a0seed, a1seed, None, None, x.is_scaled)
```
mortendahl: this worries me :D don't like optional values if we can avoid it

mortendahl: keep in mind that we're building a graph, so you're defining computations, not running them

tests/test_secure_random.py
justin1121: just leftover from an old thing I was testing, will remove!

Makefile
```diff
@@ -74,7 +74,7 @@ test: lint pythoncheck
 	python examples/pond-simple.py
 	python examples/securenn-playground.py
 	python examples/federated-average/run.py
-	python -m unittest discover
+	python -m unittest discover -f
```
mortendahl: should this be default on CI?

tf_encrypted/protocol/pond.py
```diff
@@ -3377,33 +3387,33 @@ def _mask_private(prot: Pond, x: PondPrivateTensor) -> PondMaskedTensor:
...
 
     with tf.name_scope("mask"):
-
         with tf.device(prot.crypto_producer.device_name):
-            a = x.backing_dtype.sample_uniform(x.shape)
-            a0, a1 = prot._share(a)
+            a0seed = seed()
+            a1seed = seed()
+
+            a0 = x.backing_dtype.sample_uniform(x.shape, a0seed)
```
mortendahl: why `sample_uniform` in one place but `seeded_tensor` in the other?

tf_encrypted/tensor/int100.py
```diff
@@ -259,6 +262,9 @@ def lift(x) -> 'Int100Tensor':
         if isinstance(x, Int100Tensor):
             return x
 
+        if isinstance(x, Int100SeededTensor):
+            return x.expand()
```
mortendahl: afaict this could get called every time the seeded tensor is used? ie expanded tensors are not cached?

mortendahl:
a few things about naming that could lead to confusions, otherwise okay 👍

Makefile
justin1121: In my opinion yes! Early fail if anything goes wrong so you don't have to wait the full test run to find out your build has failed.

mortendahl: alright!

tf_encrypted/protocol/pond.py
justin1121: Yeah I think there's something wrong about my implementation!

justin1121: This should just be seeded_tensor. This will also fix some of the naming issues.

tf_encrypted/tensor/int100.py
justin1121: Yeah I think this needs to removed and explicitly expand in the mask plus mul functions. We can definitely make this better in the future with backing tensor caching.

justin1121: Well... maybe it can stay there for the mul function but I think if we explicitly expand inside the mask function we can cache the expanded tensors in the PondMaskedTensor.

PR #348, opened 2018-12-05
Replaced Tensorflow deprecated Printf's

### LINKED ISSUE #311: Replace all tf.Print with tf.print
Former is [deprecated per 20/8](https://www.tensorflow.org/api_docs/python/tf/Print)!

Closed by #348

kamathhrishi:
replaced the existing tf.Print() statement with tf.print(). Replaced a few of them , more to go which I will address in this PR. 

Closes #311

### DIFF
--- a/examples/bench_conv2d_sigmoid.py
+++ b/examples/bench_conv2d_sigmoid.py
@@ -156,11 +156,11 @@ else:
     def provide_input_prediction() -> tf.Tensor:
         x = tf.random_normal(shape=input_shape, dtype=tf.float32)
         tf.print(x, [x], message="x:")
-        return x
+        return w
 
     def receive_output(tensor: tf.Tensor) -> tf.Operation:
         tf.print(tensor, [tensor, tf.shape(tensor)], message="output:")
-        return tensor
+        return w
 
     with tfe.protocol.Pond(*config.get_players('server0, server1, crypto-producer')) as prot:
 
--- a/examples/bench_conv2d_sigmoid.py
+++ b/examples/bench_conv2d_sigmoid.py
@@ -65,82 +65,82 @@ else:
 
     def provide_input_conv11weights() -> tf.Tensor:
         w = tf.random_normal(shape=conv11_fshape, dtype=tf.float32)
-        return tf.Print(w, [w], message="w11:")
+        return tf.print(w, [w], message="w11:")
 
     def provide_input_conv12weights() -> tf.Tensor:
         w = tf.random_normal(shape=conv12_fshape, dtype=tf.float32)
-        return tf.Print(w, [w], message="w12:")
+        return tf.print(w, [w], message="w12:")
 
     def provide_input_pool1weights() -> tf.Tensor:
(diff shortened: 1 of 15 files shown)


examples/bench_conv2d_sigmoid.py
```diff
@@ -65,82 +65,82 @@
 
     def provide_input_conv11weights() -> tf.Tensor:
         w = tf.random_normal(shape=conv11_fshape, dtype=tf.float32)
-        return tf.Print(w, [w], message="w11:")
+        return tf.print(w, [w], message="w11:")
```
bendecoste: These will need to be changed a bit because `tf.print` doesn't return anything.

`tf.Print` returned a tensor that would put a printing op into the graph but otherwise the tensor worked the same. Since `tf.print` returns nothing we will still want to return `w` above

```python
w = tf.random_normal(shape=conv11_fshape, dtype=tf.float32)
tf.print(w, [w], message="w11:")
return w
```

Likewise for everything below

kamathhrishi: alright , I will make the changes.

kamathhrishi: I will also think about whether we need to continue support for 1.9 and comment on the issue.

justin1121:
We now only support tf 1.12+ so this PR should be fine once @bendecoste's comments are addressed.

bendecoste:
> Should this PR still be ON?

yes @kamathhrishi, we will still accept this PR if you address the comment from above: https://github.com/mortendahl/tf-encrypted/pull/348#discussion_r239517303

kamathhrishi:
> > Should this PR still be ON?
> 
> yes @kamathhrishi, we will still accept this PR if you address the comment from above: [#348 (comment)](https://github.com/mortendahl/tf-encrypted/pull/348#discussion_r239517303)

@bendecoste Working on that :)

examples/bench_conv2d_sigmoid.py
kamathhrishi: > These will need to be changed a bit because `tf.print` doesn't return anything.
> 
> `tf.Print` returned a tensor that would put a printing op into the graph but otherwise the tensor worked the same. Since `tf.print` returns nothing we will still want to return `w` above
> 
> ```python
> w = tf.random_normal(shape=conv11_fshape, dtype=tf.float32)
> tf.print(w, [w], message="w11:")
> return w
> ```
> 
> Likewise for everything below

@bendecoste Sorry for the delay , was busy finishing the semester. I have made a few changes , let me know if those changes were done right. If , yes I will continue to cover rest of the codebase.

kamathhrishi: Do I have to modify the .rst files as well?

bendecoste:
ideally yes

kamathhrishi:
I think I have made some mistakes , I will correct them. Just to clarify. 

`return tf.Print([], [average], summarize=10, message="Average:")`

should be 

`tf.print([], [average], summarize=10, message="Average:")
 return([])
`

or

`tf.print([], [average], summarize=10, message="Average:")
 return([average])
`

PR #352, opened 2018-12-11
Required space to batch paddings converter

yanndupis:
Hey @jvmancuso,

I think this pr is ready to be reviewed / optimized. 

Thanks,


tf_encrypted/convert/register.py
```diff
@@ -386,14 +385,51 @@ def space_to_batch_nd(converter, node, inputs):
...
 
+def required_space_to_batch_paddings(converter: Converter, node: Any, inputs: List[str]):
+
+    inputs_node = [converter.outputs[inputs[i]] for i in range(len(inputs))]
+    inputs_int32 = []
+    for i in range(len(inputs_node)):
+        if isinstance(inputs_node[i], tf.NodeDef):
+            inputs_int32.append(nodef_to_numpy_array(inputs_node[i]))
+        else:
+            inputs_int32.append(tf.cast(inputs_node[i].reveal().decode(), tf.int32))
```
mortendahl: to avoid revealing anything by accident, we should probably either fail or print warning

tests/test_convert.py
```diff
@@ -43,9 +43,22 @@ def _assert_successful_conversion(prot, graph_def, actual, *input_fns, **kwargs)
...
+            if isinstance(x, (list, tuple)):
+                assert isinstance(actual, (list, tuple)), "expected output to be tensor sequence"
+                try:
+                    output = sess.run([xi.reveal() for xi in x], tag='reveal')
+                except AttributeError:
+                    # assume all xi are all public
+                    output = sess.run([xi for xi in x], tag='reveal')
+                for o_i, a_i in zip(output, actual):
+                    np.testing.assert_array_almost_equal(o_i, a_i, decimal=3)
+            else:
```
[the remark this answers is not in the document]
jvmancuso: good point!

tf_encrypted/convert/register.py
jvmancuso: agreed, I'd been wondering about this but had deferred it in the hopes that this op would only ever encounter public data.  will patch with a warning!

mortendahl: as discussed with @yanndupis this should only happen during testing but better to play it safe

justin1121: Hmmm anyway we can just form the test so this doesn't happen and remove this else clause?

mortendahl: agree if possible! 👍

jvmancuso: Seems to me that the problem is that the converter assumes that all placeholders should be private.  It's likely this will only happen during testing for this specific op, but it points to a larger problem with the converter, since there may be graphs where we'd want placeholders to the graph to be a mix of public/private values.

jvmancuso: Hard to come up with an exhaustive list, but this op is a concrete example of when neither party "owns" the Placeholder input in a privacy sense.  It comes from the model-owner, but it's not a private value, since both parties will eventually know what those are based on the shapes of other tensors passing through the graph.  More generally, whenever we have a Placeholder that represents a Shape, we'd probably want that to be a public input in TFE.

mortendahl:
lgtm 👍

tf_encrypted/convert/register.py
justin1121: Yep makes sense to me!

Issue #345, opened 2018-12-04
Tensorflow version support


justin1121:
We implicitly support versions 1.9, 1.10, 1.11 and 1.12 of tensorflow but we're not testing against all of these versions continuously, only 1.9 (https://github.com/mortendahl/tf-encrypted/issues/307). We need to decide whether we'd like to support all of these versions or perhaps only the current latest version or perhaps even latest release plus the previous release.

Some things to consider:

1. Tensorflow is rapidly improving and by supporting all the way back to 1.9 we might be losing out on some improvements in the API that we can't take advantage of because we need to support backward compatibility.
2. Tensorflow 2.0 will be coming out soon (Q1 2019??) and its unknown if it will be even feasible to support this along with all the 1.x versions. There will be a tensorflow module that points to all the 1.x functions but I think if we were to  upgrade to 2.0 we'd want to go all in.
3. It would be interesting to do some investigation to find out what is actually being used out in the wild.
4. We haven't run into any critical bugs (hard to fix bugs between tf version) but it could potentially happen in the future as we continue to support all these versions.

justin1121:
More arguments for ditching 1.11.0 and younger: 

https://github.com/mortendahl/tf-encrypted/issues/270#issuecomment-444646598

https://github.com/mortendahl/tf-encrypted/issues/311

bendecoste:
+1 I say we can 1.11 and younger.  The project is young enough that we shouldn't be bogging ourselves down with legacy support.  If no one objects we can formalize this in the README/CI/wherever else that needs to happen to close out this issue.

justin1121:
I think I meant 1.11 and older 👅

kamathhrishi:
I guess we should test for a later version keeping in mind the time we would take for more secure tfencrypted version ? I am sorry , I won't be able to analyze the library from a production level point of view. But , my general opinion.

bendecoste:
re https://github.com/mortendahl/tf-encrypted/issues/311#issuecomment-446160727 let's remove 11 and older.

What actually needs to be done?

- [ ] update to readme
- [ ] updates to CI?
- [ ] update to `setup.py`

anything else?

justin1121:
Sounds about right to me!

PR #354, opened 2018-12-13
only support tensorflow 1.12

justin1121:
Closes https://github.com/mortendahl/tf-encrypted/issues/345

This removes the implicit support we had for other versions of tensorflow. Only support 1.12 now!


ianlivingstone:
We should document the tensorflow version we support in the README and inside the installation instructions we have written in the docs, just so its' explicit for anyone looking to pickup and use the package

PR #359, opened 2018-12-19
avoid warnings during tests

mortendahl:
this test was causing some unjust warnings; setting log level higher for this scope removes them

### DIFF
--- a/tests/test_convert.py
+++ b/tests/test_convert.py
@@ -1,15 +1,14 @@
-import unittest
-import os
-import logging
-
 from typing import List, Tuple
+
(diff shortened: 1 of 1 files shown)


ianlivingstone:
Why is the warning being generated? Is it an issue with the test? Perhaps the test needs to be rewritten? Is the warning coming from us or tensorflow?

mortendahl:
Ah maybe you're not the one I had the discussion with 🙂 added the warning
since you're not normally supposed to do what the test is doing, however
for the test it's fine

PR #360, opened 2018-12-19
expansion doesn't need to be stateful

mortendahl:
blocked by #361


mortendahl:
actually, TF seems to be too "smart" for this, and will transfer expanded tensors instead of seeds if expanders are stateless; blocking until we understand TF internals better

mortendahl:
moved to issue #362

PR #369, opened 2018-12-20
Debug tcn

yanndupis:
Hey guys,

Here is a PR for the conv1D stuff. 

It handles several edge cased in the converter that where missing. The Dense layer supports transpose for the various inputs. Expand dim has been fixed. It wasn't grabbing correctly the axis input on the graph. We also decided not making the zero tensor explicit. 

Thank you!


tf_encrypted/convert/register.py
```diff
@@ -125,6 +136,12 @@ def sigmoid(converter: Converter, node: Any, inputs: List[str]) -> Any:
 
 def strided_slice(converter: Converter, node: Any, inputs: List[str]) -> Any:
     input = converter.outputs[inputs[0]]
+
+    if isinstance(input, tf.NodeDef):
```
mortendahl: what can `input` be here if not `NodeDef`?

tf_encrypted/convert/register.py
```diff
@@ -363,6 +392,7 @@ def avgpool(converter: Converter, node: Any, inputs: List[str]) -> Any:
 
 
 def concat(converter: Converter, node: Any, inputs: List[str]) -> Any:
+    print("input list", inputs)
```
mortendahl: leftover?

tf_encrypted/convert/register.py
```diff
@@ -441,6 +471,8 @@ def nodef_to_public_pond(converter, x):
             nums = x.attr["value"].tensor.float_val
         elif dtype == tf.float64:
             nums = x.attr["value"].tensor.float_val
+        elif dtype == tf.int32:
```
mortendahl: should `int32` be supported above as well?

tf_encrypted/layers/dense.py
```diff
@@ -37,25 +40,39 @@ def initialize(
...
 
         self.weights = self.prot.define_private_variable(initial_weights)
-        self.bias = self.prot.define_private_variable(initial_bias)
 
     def forward(self, x):
         self.layer_input = x
-        y = x.matmul(self.weights) + self.bias
+
+        if self.transpose_input:
+            self.layer_input = self.layer_input.transpose()
```
mortendahl: would this transpose more than once if `forward` is called repeatedly?

tf_encrypted/layers/dense.py
```diff
@@ -37,25 +40,39 @@ def initialize(
...
 
         d_weights = x.transpose().matmul(d_y)
-        d_bias = d_y.reduce_sum(axis=0)
-
         self.weights.assign((d_weights * learning_rate).neg() + self.weights)
-        self.bias.assign((d_bias * learning_rate).neg() + self.bias)
+
+        if self.bias:
+            d_bias = d_y.reduce_sum(axis=0)
+            self.bias.assign((d_bias * learning_rate).neg() + self.bias)
```
mortendahl: doesn't `assign` return an operation that has to be called for it to have any effect?

tf_encrypted/convert/register.py
yanndupis: if it's not a NodeDef it would be already a private pond tensor.

tf_encrypted/convert/register.py
yanndupis: not sure, I understand. did you mean `int64`?

tf_encrypted/layers/dense.py
yanndupis: I think we should be good because every time we run forward, self.layer_input will be assign first to `x`, which is the forward input. However, we will have an issue for the weight matrix. I think we could move this weight transpose to the initialize phase.

tf_encrypted/layers/dense.py
yanndupis: hmm not sure, it was like that when Morgan was training the logistic regression and it seemed to work.

jvmancuso: yes it does: https://github.com/mortendahl/tf-encrypted/blob/3a55ca33247dad91199669a72c3e51511f7b698c/tf_encrypted/protocol/pond.py#L614

yanndupis:
Hey @mortendahl , I think all the comments should be addressed. Thanks for your review!

tf_encrypted/layers/dense.py
jvmancuso: why wouldn't we do `self.bias -= (d_bias * learning_rate)` here?

yanndupis: Thanks for your comment @jvmancuso let me add this

Issue #248, opened 2018-10-17
Bug in network B and C 


yanndupis:
[Network A](https://github.com/mortendahl/tf-encrypted/blob/master/examples/securenn/network_a.py) works perfectly fine since we are using the exact `ReLU` with `securenn` and improved the tfe api. However we still get a very bad accuracy for [network b](https://github.com/mortendahl/tf-encrypted/blob/master/examples/securenn/network_b.py) and [c](https://github.com/mortendahl/tf-encrypted/blob/master/examples/securenn/network_c.py). We are aware that we have a bug with the exact `ReLU` that needs to be fixed. However, not sure it's the cause of the bad accuracy. When I used the `.pb` file approach for network c last week, I was getting a good accuracy on the mnist set.

ianlivingstone:
Does this block anything we need for the launch @yanndupis ?

yanndupis:
I don't think it does. As an example we will show network a. Thanks.

ianlivingstone:
Did we track down the bug?

yanndupis:
The bug hasn't been tracked. Let's come back to it once `securenn + int64 + exact relu` is completed.

mortendahl:
@yanndupis can this be closed?

ianlivingstone:
cc @yanndupis ^

yanndupis:
I will check this week if with the securenn protocol completed now it's working.

mortendahl:
any updates @yanndupis?

yanndupis:
We found the bugs in for this issue. There were missing relus, and we were using channel last during training for the convolution and pooling layers however we were using channel first for secure inference. Once the PR #375 is merged we can close this issue.

PR #375, opened 2019-01-04
Securenn examples fixed

yanndupis:
Hello,

With this pr, the securenn network b and c are now working.  There was a missing relu, and we were using channel last during training for the convolution and pooling layers however we were using channel first for secure inference. 

For plaintext training, I wanted to use channel first in Tensorflow. However I was getting the following error even if I was training on CPU: `Default AvgPoolingOp only supports NHWC on device type CPU`. 

So instead I did channel last for the secure inference part. I used the layer object that gives the option to specify channel last for the convolution and pooling layers instead of using the primitives. 

Thank you!

fixe #248

### DIFF
--- a/examples/securenn/network_c.py
+++ b/examples/securenn/network_c.py
@@ -169,14 +169,9 @@ class PredictionClient():
             prediction = tf.argmax(likelihoods, axis=1)
             eq_values = tf.equal(prediction, tf.cast(y_true, tf.int64))
             acc = tf.reduce_mean(tf.cast(eq_values, tf.float32))
+            op = tf.print('Expected:', y_true, '\nActual:', prediction, '\nAccuracy:', acc)
 
-            expect_out = tf.print("EXPECT: ", y_true, summarize=self.BATCH_SIZE)
-
-            actual_out = tf.print("ACTUAL: ", prediction, summarize=self.BATCH_SIZE)
-        
-            accuracy_out = tf.print("Acuraccy: ", acc, summarize=self.BATCH_SIZE)
-      
-            return [expect_out, actual_out, accuracy_out]
+            return op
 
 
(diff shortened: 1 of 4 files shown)


examples/securenn/network_c.py
```diff
@@ -116,8 +113,7 @@ def loop_body(i: tf.Tensor, max_iter: tf.Tensor, nb_epochs: tf.Tensor, avg_loss:
             layer2 = pooling(tf.nn.relu(conv2d(layer1, Wconv2, self.STRIDE) + bconv2))
             layer2 = tf.reshape(layer2, [-1, self.HIDDEN_FC1])
             layer3 = tf.nn.relu(tf.matmul(layer2, Wfc1) + bfc1)
-            layer4 = tf.nn.relu(tf.matmul(layer3, Wfc2) + bfc2)
-            logits = tf.matmul(layer4, Wfc3) + bfc3
+            logits = tf.matmul(layer3, Wfc2) + bfc2
```
jvmancuso: any reason we've removed the penultimate layer here?  [this diagram](https://www.google.com/url?sa=i&source=images&cd=&cad=rja&uact=8&ved=2ahUKEwieqbngr97fAhXCTd8KHYkpBX4QjRx6BAgBEAU&url=https%3A%2F%2Fwww.researchgate.net%2Ffigure%2FThe-LeNet-5-Architecture-a-convolutional-neural-network_fig4_321586653&psig=AOvVaw0NjyTeTVArGQsqyznFd80v&ust=1547043357750404) suggests the network should have 2 linear layers before the output layer

jvmancuso: ah, just checked the SecureNN paper -- looks like they removed the penultimate layer from LeNet, nvm :)

PR #379, opened 2019-01-08
Improve examples and documentation

### LINKED ISSUE #126: Add public division to POND

### LINKED ISSUE #134: federated-average example is quite vague
This is probably the first thing that will be ran by a lot of people.  The output of this example is three lines that say "Average" and then a vector.  It's not very clear what this is doing or what is going on.  I think with some better logging it could become more obvious.

### LINKED ISSUE #180: Make config.json files super simple
Make config.json files super simple so they can intuitively be created by hand, e.g. just a hostmap. If needed then remove need for supporting LocalConfig in config.json.

cc @morgangiraud happy for any inputs you may have!

### LINKED ISSUE #333: Proper logging
Introduce [logging](https://docs.python.org/3/library/logging.html) mechanism so we can better issue warnings, errors, etc

- [x] log to `tf_encrypted` instead of the current `root`
- [ ] a way of turning warnings into fatals for production (cc @justin1121)

### LINKED ISSUE #373: Running instructions for multiple machines (non cloud based)
Hi,

I can understand why you use GCP in the instructions since not many people have multiple machines and VMs are not desired for performance reasons. However, there are those of us who would like to try it on multiple machines outside the cloud environment on bare metal servers. 

Any chance you could amend the instructions o include such scenario?

Thanks,

### LINKED ISSUE #391: Consistent filenames
Some of the files in this repo use dashes (`-`) and some use underscores (`_`), we should use the same each time imo

e.g. 

```
examples/mnist_deep_cnn.py
examples/securenn-playground.py
examples/federated-average/
examples/house_credit_default/
tf_encrypted/
```

### LINKED ISSUE #404: Use "TF Encrypted" instead of "tf-encrypted" in all texts

mortendahl:
- [x] tighten `examples` directory (move to `deprecated` when in doubt)
- [x] remove examples from `make test`
- [x] running documentation
- [x] MNIST readme: update
- [x] docs readme: only give quick overview
- [x] proper figures for examples
- [x] fix securenn networks

closes #391 
closes #126 
closes #404 
closes #180
closes #134 
closes #333 
closes #373


tf_encrypted/config.py
```diff
@@ -82,306 +76,210 @@ def get_tf_config(self) -> Tuple[str, tf.ConfigProto]:
...
 
-    def get_players(self, names: Union[List[str], str]) -> List[Player]:
+    def get_players(self, names):
         if isinstance(names, str):
             names = [name.strip() for name in names.split(',')]
         assert isinstance(names, list)
         return [player for name, player in self._players.items() if name in names]
 
-    def server(self, name: str) -> tf.train.Server:
+    def server(self, name, start=True):
```
justin1121: Just curious if there's somewhere where we'd want to not start the server on creation?

justin1121:
Tried running the examples and had some problem with the logreg example and the neural networks examples. 

With logreg there was a problem with running `tf.print` perhaps we should just use `tf.Print` and that should work.

As for the neural networks example its not clear that you need to use the data from the mnist example. Perhaps we should just copy the download script there as well.

justin1121:
Same `tf.print` problem with netword_d.py.

tf_encrypted/config.py
mortendahl: I'm not sure no. My reasoning was that the TF server takes this as a parameter, and `python -m tf_encrypted.player` makes an assumption that start is true; so if this were to ever change our launching of servers would be broken.

examples/simple-average/README.md
```diff
@@ -0,0 +1,105 @@
+
+# Overview
+
+This example illustrates how TF Encrypted can be used to securely compute the average of inputs from a set of input providers, with the intend of providing a simple starting point.
```
jvmancuso: intend --> intent or intention

docs/RUNNING.md
```diff
@@ -0,0 +1,93 @@
...
+
+This document describes how to run secure computations using TF Encrypted. There are two ways of doing so as detailed below:
+
+- Using a `LocalConfig` object: this simulates the players on the local machine using different threads, and as such is completely insecure; however, it is very convenient for testing purposes
+
+- Using a `RemoteConfig` object: this orchestrates a distributed computation between ordinary TensorFlow servers, typically running on distinct machines in a network; this is secure and represents how computations should actually be run
+
+New configurations are specified using `tfe.set_config()`. The default is `LocalConfig()`.
+
+See also the [installation instructions](/docs/INSTALL.md) for more information on installing TF Encrypted, and [our GCP documentation](/docs/GCP.md) for more instructions using it on the Google Cloud Platform.
```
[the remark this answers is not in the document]
mortendahl: well spotted!

PR #381, opened 2019-01-09
Tensor cleanup, seeding, and more

### LINKED ISSUE #366: More aggressive use of seeds
use seeded tensors where possible in SecureNN:
 - [x] int64
 - [x] int32
 - [x] odd
 - [x] prime
 - [x] int100

Child of #364

### LINKED ISSUE #380: Fix use of seeds in CRT tensors
Right now the same seed is used for each component in a CRT tensor, which is insecure.

Simple idea could be to simply use a list of seeds instead.

More involved idea could be to derive list of seeds of initial seed.

mortendahl:
Closes #366 
Closes #380

Most important changes:
- interface and behaviour of tensors made consistent
- unified int32, int64, and prime tensors, reducing code duplications
- inline code previously in `crt.py` into a new CRT tensor
- improve usage of seeded tensors
- fixed security issue in SecureNN where some operations where happening on the wrong device
- fixed security issue in int100 where some seeds where reused

**_Note for review: commits represent logical progression and it may be easier to go commit by commit instead of looking at all changes at once._**


tf_encrypted/protocol/securenn/odd_tensor.py
```diff
@@ -58,170 +63,225 @@ def modulus(self):
...
-        raise NotImplementedError()
+        return OddDenseTensor(z, self.factory)
 
-    def im2col(self, h_filter, w_filter, padding, strides) -> 'OddImplicitTensor':
-        raise NotImplementedError()
-
-    def conv2d(self, other, strides, padding='SAME') -> 'OddImplicitTensor':
-        raise NotImplementedError()
+    def bits(self, factory=None):
+        if factory is None:
```
justin1121: Small simplification:

```
if factory is None:
  factory = self.factory

return factory.tensor(binarize(self.value))

File: tf_encrypted/tensor/native.py [hunk 2]
```diff
@@ -167,26 +197,42 @@ def __getitem__(self, slice):
...
+            return DenseTensor(value)
 
         def im2col(self, h_filter: int, w_filter: int, padding: str, strides: int):
             return DenseTensor(im2col(self.value, h_filter, w_filter, padding, strides))
 
         def conv2d(self, other, strides: int, padding: str = 'SAME'):
-            x, y = _lift(self, other)
-            return conv2d(x, y, strides, padding)  # type: ignore
+            if EXPLICIT_MODULUS is not None:
+                # TODO(Morten) any good reason this wasn't implemented for PrimeTensor?
```
justin1121: I just don't think it was done!

tf_encrypted/protocol/securenn/odd_tensor.py
```diff
@@ -12,283 +12,276 @@
...
-    if isinstance(y, OddTensor):
+            if isinstance(x, int):
+                return y.factory.tensor(np.array([x])), y
 
-        if isinstance(x, int):
-            return y.factory.tensor(np.array([x])), y
+        raise TypeError("Don't know how to lift {} {}".format(type(x), type(y)))
 
-    raise TypeError("Don't know how to lift {} {}".format(type(x), type(y)))
+    def _lessthan_as_unsigned(x, y, bitlength):
```
justin1121: wonder if we should put this under its own scope?

tf_encrypted/protocol/securenn/odd_tensor.py
justin1121: Noticed somewhere you mentioned direct constructor is better at compile time so maybe this isn't relevant.

justin1121: Also not even sure this function still exists 😄 !

mortendahl: in this case it's actually an optimization to call the constructor directly, since it avoids the check for -1 values in the factory! works here since we know `binarize` will not return -1 values and hence no need to correct for these.

PR #382, opened 2019-01-09
Fix multiple placeholders

yanndupis:
Hello, 

This PR fix the multiple placeholder issue and there was also a missing edge case for slice.

Thanks,

### DIFF
--- a/bin/run
+++ b/bin/run
@@ -81,14 +81,14 @@ with gfile.FastGFile(model_file, 'rb') as f:
 
 inputs = []
 for i, spec in enumerate(input_spec):
-    def scope(i: int) -> Callable[[], tf.Tensor]:
+    def scope(i: int, spec: dict) -> Callable[[], tf.Tensor]:
         def provide_input() -> tf.Tensor:
             pl = tf.placeholder(tf.float32, shape=spec['shape'], name="api/{}".format(i))
             return pl
 
         return provide_input
 
-    inputs.append(scope(i))
+    inputs.append(scope(i, spec))
 
 c = tfe.convert.convert.Converter()
 y = c.convert(graph_def, tfe.convert.register(), 'input-provider', inputs)
--- a/tf_encrypted/convert/register.py
+++ b/tf_encrypted/convert/register.py
@@ -472,11 +472,11 @@ def slice(converter, node, inputs):
     else:
         input_out = input
 
-    # Slice is a special case of strided_slice. Slice takes size (the number of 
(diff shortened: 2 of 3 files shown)


bin/run
```diff
@@ -81,14 +81,14 @@ with gfile.FastGFile(model_file, 'rb') as f:
 
 inputs = []
 for i, spec in enumerate(input_spec):
-    def scope(i: int) -> Callable[[], tf.Tensor]:
+    def scope(i: int, spec: dict) -> Callable[[], tf.Tensor]:
```
mortendahl: I don't see why you need to pass `spec` here; did you run into trouble without it?

mortendahl: likewise, you shouldn't need to pass `i`

yanndupis: It's not exactly clear why I was running into trouble without `spec` in scope but it solved my problem. 

Let's say I have two placeholders `input_char` and `eff_history`:
`input_spec = [{'name': 'input_chars', 'dtype': 1, 'shape': [1, 400, 100]}, {'name': 'eff_history', 'dtype': 3, 'shape': [1]}]`

When I was doing `print(spec)`, after line 83, I was getting correctly the two placeholder `{'name': 'input_chars', 'dtype': 1, 'shape': [1, 400, 100]}` and `{'name': 'eff_history', 'dtype': 3, 'shape': [1]}`.

However when I was doing `print(spec)` after line 85, within `provide_input` I was getting two times {'name': 'eff_history', 'dtype': 3, 'shape': [1]} instead of getting `input_char` and `eff_history`.

When I pass `spec` to `scope` everything works properly.

mortendahl: you're right, the below prints `4` six times! ([python is silly](https://eev.ee/blog/2011/04/24/gotcha-python-scoping-closures/))

good catch, forget what I said

```python
functions = []

for i in range(5):
  def function():
    return i
  functions.append(function)

print(i)

for f in functions:
  print(f())
```

tf_encrypted/convert/register.py
```diff
@@ -472,7 +472,15 @@ def slice(converter, node, inputs):
     else:
         input_out = input
 
-    end = begin + size
+    end = np.zeros(len(begin))
+    input_shape = input.shape.as_list()
+
+    # if size is negative take the input dimension
```
jvmancuso: This comment could be improved :)

yanndupis: good point :)

Issue #376, opened 2019-01-07
TypeError: Forward references must evaluate to types. Got tf.int64.


mkskeller:
Despite installing the wheel linked to in the readme, I get the following error on Ubuntu 16.04 (on Ubuntu 18.04 the wheel would install at all):
```
Traceback (most recent call last):
  File "./examples/mnist/run.py", line 5, in <module>
    import tf_encrypted as tfe
  File "/home/marcel/.local/lib/python3.5/site-packages/tf_encrypted/__init__.py", line 7, in <module>
    from .session import Session, setTFEDebugFlag, setMonitorStatsFlag, setTFETraceFlag
  File "/home/marcel/.local/lib/python3.5/site-packages/tf_encrypted/session.py", line 11, in <module>
    from .protocol.pond import PondPublicTensor
  File "/home/marcel/.local/lib/python3.5/site-packages/tf_encrypted/protocol/__init__.py", line 5, in <module>
    from .securenn import SecureNN
  File "/home/marcel/.local/lib/python3.5/site-packages/tf_encrypted/protocol/securenn/__init__.py", line 3, in <module>
    from .securenn import SecureNN
  File "/home/marcel/.local/lib/python3.5/site-packages/tf_encrypted/protocol/securenn/securenn.py", line 10, in <module>
    from .odd_tensor import oddInt64factory
  File "/home/marcel/.local/lib/python3.5/site-packages/tf_encrypted/protocol/securenn/odd_tensor.py", line 14, in <module>
    class OddImplicitFactory:
  File "/home/marcel/.local/lib/python3.5/site-packages/tf_encrypted/protocol/securenn/odd_tensor.py", line 16, in OddImplicitFactory
    def __init__(self, native_type: Union[Type['tf.int32'], Type['tf.int64']]) -> None:
  File "/usr/lib/python3.5/typing.py", line 552, in __getitem__
    dict(self.__dict__), parameters, _root=True)
  File "/usr/lib/python3.5/typing.py", line 512, in __new__
    for t2 in all_params - {t1} if not isinstance(t2, TypeVar)):
  File "/usr/lib/python3.5/typing.py", line 512, in <genexpr>
    for t2 in all_params - {t1} if not isinstance(t2, TypeVar)):
  File "/usr/lib/python3.5/typing.py", line 1059, in __subclasscheck__
    if not issubclass(p_cls, p_self):
  File "/usr/lib/python3.5/typing.py", line 190, in __subclasscheck__
[...]

PR #383, opened 2019-01-10
Make custom ops optional

### LINKED ISSUE #334: Allow TFE to run without first building custom ops?
Instead of panicking if the custom ops are not available we could raise a warning (or even non-fatal error) and either reduce functionality or fall back to TF's implementation. This might give better experience than being force to compile first.

Depends on #333.

mortendahl:
Changes making custom ops (`secure_random`) optional, with warnings if not used.

Depends on #381 
Needed by #376 
Closes #334

### DIFF
--- a/tf_encrypted/protocol/pond.py
+++ b/tf_encrypted/protocol/pond.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 from typing import Tuple, List, Union, Optional, Any, NewType, Callable
 import abc
 import sys
+import logging
 from math import log2, ceil
 from random import random
 
@@ -70,6 +71,10 @@ class Pond(Protocol):
             if tensorflow_supports_int64():
                 tensor_factory = int64factory
             else:
+                logging.warning(
+                    "Falling back to using int100 tensors due to lack of int64 support. "
+                    "Performance may be improved by installing a version of TensorFlow "
+                    "supporting this (1.13+ or custom build).")
                 tensor_factory = int100factory
 
         if fixedpoint_config is None:
--- a/tf_encrypted/protocol/securenn/odd_tensor.py
+++ b/tf_encrypted/protocol/securenn/odd_tensor.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 from typing import Tuple, Optional
+from functools import partial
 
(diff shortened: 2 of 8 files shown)


tf_encrypted/protocol/pond.py
```diff
@@ -70,6 +71,9 @@ def __init__(
             if tensorflow_supports_int64():
                 tensor_factory = int64factory
             else:
+                logging.warning(
+                    "Falling back to using int100 tensors due to lack of int64 support. "
+                    "Performance may be improved by installing the latest version of TensorFlow.")
```
justin1121: hmmm kind of misleading, maybe we should link to our documentation on how to install our tensorflow custom builds? tf 1.13 should have int64 support but its not out quite yet, hopefully soon!

tf_encrypted/protocol/securenn/odd_tensor.py
```diff
@@ -70,7 +71,20 @@ def sample_uniform(self,
                            maxval: Optional[int] = None):
             assert minval is None
             assert maxval is None
-            return OddUniformTensor(shape=shape, seed=seed())
+
+            if secure_random.supports_seeded_randomness():
+                seed = secure_random.seed()
+                return OddUniformTensor(shape=shape, seed=seed)
+
+            elif secure_random.supports_secure_randomness():
```
justin1121: Not sure I get the point of having these two flags. They both come from the same check for the secure random module so they are both going to be false all the time.

tf_encrypted/protocol/pond.py
mortendahl: "by installing a version of TensorFlow supporting this (custom build or 1.13+)."?

justin1121: Yeah sounds good!

tf_encrypted/protocol/securenn/odd_tensor.py
mortendahl: i basically just wanted to have an easy way to turn off seeding in upcoming benchmarks

justin1121: Ahh yes good idea!

PR #384, opened 2019-01-10
Conv2d - fix strides input - converter

yanndupis:
Hey @jvmancuso , 

There was an issue with the strides input for the conv2d layer. For example for stride = 2, the strides input would look like `[1, 2, 2, 1]`. But we were taking only at the first element so the stride was equal to 1. 

Thanks,


yanndupis:
Thanks for your review!

PR #389, opened 2019-01-23
House Credit Default Demo!

justin1121:
Logistic regression example showing off how to run simple ML architectures privately.

### DIFF
--- a/bin/run
+++ b/bin/run
@@ -38,14 +38,10 @@ if config.protocol_name == 'securenn':
     tfe.set_protocol(SecureNN())
 else:
     tfe.set_protocol(Pond())
-    
-model_file = os.path.join(
-    os.path.dirname(os.path.realpath(__file__)),
-    "..",
-    "models",
-    "{}.pb".format(config.model_name),
-)
+
+model_file = config.model_name
 model_path = Path(model_file)
+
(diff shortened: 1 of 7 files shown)


bin/run
```diff
@@ -64,7 +60,7 @@ if not model_path.is_file() and config.model_url is not '':
 tf.reset_default_graph()
 
 input_spec = []
-with gfile.FastGFile(model_file, 'rb') as f:
+with gfile.GFile(model_file, 'rb') as f:
```
bendecoste: Why did you decide to slow down gfile? 😅

examples/house_credit_default/main.py
```diff
@@ -0,0 +1,191 @@
...
+    graph = graph_util.remove_training_nodes(graph)
+    path = graph_io.write_graph(graph, ".", filename, as_text=False)
+    print('saved the frozen graph (ready for inference) at: ', path)
+
+
+def print_nodes(graph):
+    print([n.name for n in graph.as_graph_def().node])
+
+
+def data_prep():
```
bendecoste: this is also defined in utils

bin/run
justin1121: It was deprecated! Assume they defaulted to fast haha.

PR #390, opened 2019-01-24
bin/run changes

bendecoste:
Modified bin/run so you can pass --plaintext as well as
specify how many iterations you want the test to run


bin/run
```diff
@@ -113,17 +106,78 @@ else:
...
+        sess.run(tf.global_variables_initializer())
+
+        # TODO This isn't a good way to get the input/output nodes and we shouldn't
+        #      hardcode the input either!
+        print('running', runs, 'predictions')
+        for x in range(0, runs + 1): # first run is warmup
+            output = sess.graph.get_operations()[-2].values()[0]
+            input = sess.graph.get_operations()[0].values()[0]
+
+            sess.run(output, {input: np.random.rand(1000,528)}, tag='prediction')
```
justin1121: we're loading the input data from the input_file cli argument, should be able to pass input_data directly in here

PR #393, opened 2019-01-28
added secure_model to private_model

bendecoste:


tests/test_secure_model.py
```diff
@@ -0,0 +1,30 @@
...
+
+import numpy as np
+import tensorflow as tf
+from tf_encrypted.private_model import secure_model
+
+
+class TestSecureModel(unittest.TestCase):
+
+    def test_secure_model(self):
+        tf.random.set_random_seed(42)
```
bendecoste: seed is probably not necessary but i figured we would occasionally have precision errors without it.

PR #394, opened 2019-01-29
add flatten as a special op

justin1121:


tf_encrypted/convert/convert.py
```diff
@@ -184,12 +186,12 @@ def select_relevant_ops(special_ops: list,
...
 
-    for i in range(len(special_ops)):
-        for n in graph_def.node:
-            if special_ops[i] in n.name:
-                if n.name in all_special_op_inputs or n.name in all_special_op_outputs:
-                    trimmed_graph[n.name] = n
-            else:
+    for n in graph_def.node:
+        # TODO generic for n number of special_ops
+        if special_ops[0] in n.name or special_ops[1] in n.name:
```
yanndupis: Looks good. good catch, it should eliminate a loop. We should probably open an issue to make this condition make generic.

justin1121: Yep will open an issue 😄

yanndupis: Thank you :)

PR #414, opened 2019-03-27
tf.negative

yanndupis:
Hey @mortendahl,

This PR implements the [tf.negative](https://www.tensorflow.org/api_docs/python/tf/math/negative) operation (`tf.negative([-1, 0, 1]) == [1, 0, -1]`). I named it `neg` so it doesn't get mixed with the negative operation in the securenn protocol (`negative([-1, 0, 1]) == [1, 0, 0]`).  

This operation appeared in one of the convolution network when I turned off in the batchnorm layer the parameters scale and offset. 

Thanks!

### DIFF
--- a/tf_encrypted/protocol/pond.py
+++ b/tf_encrypted/protocol/pond.py
@@ -841,6 +841,28 @@ class Pond(Protocol):
         raise TypeError("Don't know how to reshape {}".format(type(x)))
 
     @memoize
+    def neg(self, x: "PondTensor"):
+        """
+        reshape(x, shape) -> PondTensor
+
+        Reshape `x` into a tensor with a new `shape`.
+
+        :param PondTensor x: Input tensor.
+        :param (int,...) shape: Shape of output tensor.
+        """
+
+        if isinstance(x, PondPublicTensor):
+            return _neg_public(self, x)
+
+        if isinstance(x, PondPrivateTensor):
+            return _neg_private(self, x)
+
+        if isinstance(x, PondMaskedTensor):
+            return _neg_masked(self, x)
+
+        raise TypeError("Don't know how to reshape {}".format(type(x)))
+
+    @memoize
     def expand_dims(self, x: "PondTensor", axis=None):
 
         if isinstance(x, PondPublicTensor):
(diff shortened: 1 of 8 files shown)


mortendahl:
hi @yanndupis, good work! I suggest we rename the existing method in SecureNN, maybe to `is_negative`, in order to stick with the TF API; anything else will only cause more confusions down the line.

tf_encrypted/protocol/pond.py
```diff
@@ -3526,6 +3556,77 @@ def _reshape_masked(
...
+        with tf.device(prot.server_0.device_name):
+            x0_neg = x0.neg()
+
+        with tf.device(prot.server_1.device_name):
+            x1_neg = x1.neg()
+
+        return PondPrivateTensor(prot, x0_neg, x1_neg, x.is_scaled)
+
+
+def _neg_masked(
```
mortendahl: this works out but just a small note of caution that this is not always straight-forward and could lead to privacy issues 👍

yanndupis:
Hey @mortendahl ,

Thank you very much for your review. 

I have realized that `negative` was already implemented on the backing tensor. 

What do you think is better in `int100.py`: 
```
def negative(self):
            # TODO[Morten] there's probably a more efficient way
            return FACTORY.zero() - self
```
 or 

```
def neg(self):
            backing = [tf.math.negative(xi) for xi in self.backing]
            return DenseTensor(backing)
```

and in `native.py`:

```
def negative(self):
            value = tf.negative(self.value)
            if EXPLICIT_MODULUS is not None:
                value %= EXPLICIT_MODULUS
            return DenseTensor(value)
``` 

or 

```
def neg(self):
            return DenseTensor(tf.negative(self.value))
```

Thank you!

mortendahl:
Hi @yanndupis, my suggestion:

- For int100:

```python
def neg(self):
    backing = [tf.negative(xi) % mi for xi, mi in zip(self.backing, MODULI)]
    return DenseTensor(backing)
```

- For native:

```python
def negative(self):
    value = tf.negative(self.value)
    if EXPLICIT_MODULUS is not None:
        value %= EXPLICIT_MODULUS
    return DenseTensor(value)
```

yanndupis:
Hey @mortendahl , thanks for your review! Your comments should be addressed with this pr. thanks.

PR #421, opened 2019-04-05
Add converter support for Split

yanndupis:
Hey @jvmancuso , 

Would you mind reviewing this PR?

Thank you!

### DIFF
--- a/tests/test_convert.py
+++ b/tests/test_convert.py
@@ -550,18 +550,16 @@ def export_squeeze(filename, input_shape):
 
 def run_split(input):
     a = tf.placeholder(tf.float32, shape=input.shape, name="input")
-    x1, x2 = tf.split(a, num_or_size_splits=2, axis=1)
-    y = tf.add(x1, x2)
+    x = tf.split(a, num_or_size_splits=2, axis=1)
     with tf.Session() as sess:
-        output = sess.run(y, feed_dict={a: input})
+        output = sess.run(x, feed_dict={a: input})
     return output
 
 
 def export_split(filename, input_shape):
     a = tf.placeholder(tf.float32, shape=input_shape, name="input")
-    x1, x2 = tf.split(a, num_or_size_splits=2, axis=1)
-    y = tf.add(x1, x2)
-    return export(y, filename)
+    x = tf.split(a, num_or_size_splits=2, axis=1)
+    return export(x, filename)
 
 
 def run_sub(input):
--- a/tests/test_convert.py
+++ b/tests/test_convert.py
@@ -50,7 +50,6 @@ class TestConvert(unittest.TestCase):
     @staticmethod
     def _assert_successful_conversion(prot, graph_def, actual, *input_fns, decimals=3, **kwargs):
         prot.clear_initializers()
-
         converter = Converter(tfe.get_config(), prot, 'model-provider')
         x = converter.convert(graph_def, register(), 'input-provider', list(input_fns))
 
@@ -147,6 +146,10 @@ class TestConvert(unittest.TestCase):
     def test_squeeze_convert(self):
         test_input = np.ones([1, 2, 3, 1])
         self._test_with_ndarray_input_fn('squeeze', test_input, protocol='Pond')
+
+    def test_split_convert(self):
+        test_input = np.ones([2, 4])
+        self._test_with_ndarray_input_fn('split', test_input, protocol='Pond')
 
     def test_sub_convert(self):
         test_input = np.ones([28, 1])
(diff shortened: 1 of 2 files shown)


tf_encrypted/convert/register.py
```diff
@@ -40,6 +40,8 @@ def register() -> Dict[str, Any]:
         'flatten': flatten,
         'Slice': slice,
         'Neg': negative,
+        'Split': split,
+        'Identity': identity,
```
jvmancuso: I'm not sure about how we should handle these Identity ops long-term, since they're possible to prune out using various graph transformation tools from TF.  Let's keep it, but note we may change it in the future.

tests/test_convert.py
```diff
@@ -545,6 +548,22 @@ def export_squeeze(filename, input_shape):
     return export(x, filename)
 
 
+def run_split(input):
+    a = tf.placeholder(tf.float32, shape=input.shape, name="input")
+    x1, x2 = tf.split(a, num_or_size_splits=2, axis=1)
+    y = tf.add(x1, x2)
```
jvmancuso: Can we remove this line?  We don't want to test Pond.add in this one, only split, but currently a bug in Pond.add will force this test case to fail.  There is a similar problem with `test_flatten` that i'm fixing in an upcoming PR, but I'd like to merge this one before I submit that.

tests/test_convert.py
```diff
@@ -545,6 +548,22 @@ def export_squeeze(filename, input_shape):
...
+    with tf.Session() as sess:
+        output = sess.run(y, feed_dict={a: input})
+    return output
+
+
+def export_split(filename, input_shape):
+    a = tf.placeholder(tf.float32, shape=input_shape, name="input")
+    x1, x2 = tf.split(a, num_or_size_splits=2, axis=1)
+    y = tf.add(x1, x2)
```
jvmancuso: same as above

tf_encrypted/convert/register.py
yanndupis: good point.

tests/test_convert.py
yanndupis: Let me remove this.

tests/test_convert.py
yanndupis: fixed

avitalsh:
Hi, I thinks there is a problem with the split converter - it returns only the first split:
https://github.com/tf-encrypted/tf-encrypted/blob/c9200f3d814f54128c4bad24c615458dfc8f253d/tf_encrypted/convert/register.py#L398

For example, I tried running 
```
tf.split(input, 3, axis=-1)
```
where input shape is (1, 10, 10, 3). The result shape is (3, 1, 10, 10, 1).
After converting the model using secure_model, the results shape for the same input is (1, 10, 10, 1).

jvmancuso:
strange,  I'm surprised that it's passing the conversion tests.

@yanndupis can you take a look at this one?

yanndupis:
yeah, if that's the problem, that's strange the test is passing. I will look into it. thanks

avitalsh:
Hi @yanndupis , did you get a chance to look into this?

yanndupis:
Hey @avitalsh , sorry for the belated response. There is definitely a problem, I will have to continue to investigate on Monday. 

The test was passing because in the graph, after the split operation, there was a `packing` operation, re-using several time the same first tensor returned by tf.split. Because the input tensor in the test was including just ones, it didn't detect this bug. 

The converter should be able to handle several outputs when we register it as special op but for some reason, it's not working. I will continue to investigate. 

As reference, here is the pb file generated by the test with an input shape of [1, 10, 10, 3].

```
node {
  name: "input"
  op: "Placeholder"
  attr {
    key: "dtype"
    value {
      type: DT_FLOAT
    }
  }
  attr {
    key: "shape"
    value {
      shape {
        dim {
          size: 1
        }
        dim {
          size: 10
        }
        dim {
          size: 10
        }
        dim {
          size: 3
        }
      }
    }
  }
}
node {
  name: "split/split_dim"
  op: "Const"
  attr {
    key: "dtype"
    value {
      type: DT_INT32
    }
  }
  attr {
    key: "value"
    value {
      tensor {
        dtype: DT_INT32
        tensor_shape {
        }
        int_val: -1
      }
    }
  }
}
node {
  name: "split"
  op: "Split"
  input: "split/split_dim"
  input: "input"
  attr {
    key: "T"
    value {
      type: DT_FLOAT
    }
  }
  attr {
    key: "num_split"
    value {
      i: 3
    }
  }
}
node {
  name: "output/input"
  op: "Pack"
  input: "split"
  input: "split:1"
  input: "split:2"
  attr {
    key: "N"
    value {
      i: 3
    }
  }
  attr {
    key: "T"
    value {
      type: DT_FLOAT
    }
  }
  attr {
    key: "axis"
    value {
      i: 0
    }
  }
}
```

yanndupis:
Hey @avitalsh, we have just merged this PR #531. I hope it can fix your problem. Thanks for raising this bug.

avitalsh:
@yanndupis Amazing, thanks!!

PR #425, opened 2019-04-10
peg tensorflow requirement to 1.x (errors w/ tf 2.0.0-alpha0)

jamslevy:
I installed tensorflow 2.0.0-alpha0 while setting up this project and that resulted in some errors using tf-encrypted like `AttributeError: module 'tensorflow' has no attribute 'Session'`. I pegged the tensorflow requirement so that until forwards compatibility is implemented running `pip3 install -e .` will ensure that a compatible version is installed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,11 +12,11 @@ setuptools.setup(
     ]},
     python_requires=">=3.5",
     install_requires=[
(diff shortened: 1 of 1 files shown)


bendecoste:
Nice add! We don't have tf 2.0 support yet :)

PR #434, opened 2019-04-17
Data-processing from saved model

yanndupis:
Hey @jvmancuso , 

With this PR we can now run on the` input-provider` device the data-preprocessing graph stored in a SavedModel file. 

Thank you!


bin/utils.py
```diff
@@ -16,14 +16,25 @@ def data_prep_from_saved_model(
     # Extract graph definition
     gdef = sess.graph_def
 
+    # Trim graph to keep only the nodes related to data pre-processing
+    # tf.graph_util.extract_sub_graph will be removed in future tf version
+    data_prep_end_node_name = data_prep_end_node.split(":")[0]
+
+    try:
```
jvmancuso: I suggest moving this try/except up into the import statements.  you can import the function directly instead of using the higher-level module syntax, which will keep the core code cleaner

yanndupis: thanks for the suggestion, let me fix that.

bin/run
```diff
@@ -81,33 +92,47 @@ with gfile.GFile(model_file, 'rb') as f:
...
-    # We will need to save our inputs as a list of np.array
-    input_data = [np.load(config.input_file)]
-    assert len(input_data) == len(input_spec), 'the input_data length ({}) should be equal the number of inputs required by the loaded model ({})'.format(
-        len(input_data), len(input_spec))
+# If data_prep_path is set then extract data-preprocessing
+# graph from the SavedModel. Finally run it on the input-provider
+if config.data_prep_path:
+    data_file_names = list_files_from_dir(config.data_dir)
+    def provide_input():
+            return data_prep_from_saved_model(config.data_prep_path,
```
mortendahl: Running a session here (inside `data_pred_from_saved_model`) is still a bit weird to me. Not blocking, the semantics just bothers me. Is it the only way you could get device placement to work?

An alternative might be to load `gdef` outside of `provide_input` and pass that to `data_pred_from_saved_model` instead of `config.data_prep_path`. Maybe this won't result in the right placement though.

bin/run
```diff
@@ -81,33 +92,47 @@ with gfile.GFile(model_file, 'rb') as f:
...
+                config.data_prep_end_node)
+
+    inputs.append(provide_input)
 else:
-    input_data = []
-    for spec in input_spec:
-        input_data.append(
-            np.random.standard_normal(spec['shape']).tolist()
+    for i, spec in enumerate(input_spec):
+        def scope(i: int, spec: dict) -> Callable[[], tf.Tensor]:
```
mortendahl: I would consider defining `scope` outside the loop, just to avoid the name hiding

bin/run
```diff
@@ -125,59 +150,87 @@ if config.plaintext:
...
-            name = "private-input_{}/api/{}:0".format(i, i)
-        pls.append(
-            tf.get_default_graph().get_tensor_by_name(name)
-        )
+    if config.data_prep_path:
+        with tfe.Session() as sess:
+            print("initing!!!")
+            sess.run(tf.global_variables_initializer())
+
+            print("warming up")
```
mortendahl: Consider using a proper logger instead of printing

bin/utils.py
```diff
@@ -0,0 +1,49 @@
...
+                                    return_elements=[data_prep_end_node])
+
+    # TFE expects tensors with fully defined shape
+    fixed_shape = [batch_size] + data_out.get_shape().as_list()[1:]
+    data_out = tf.reshape(data_out, fixed_shape)
+    return data_out
+
+
+def list_files_from_dir(directory):
+    file_names_list = tf.gfile.ListDirectory(directory)
```
mortendahl: Consider using `tf.gfile.io.listdir` instead -- seems to map to v2 of the functionality: https://github.com/tensorflow/tensorflow/blob/r1.13/tensorflow/python/lib/io/file_io.py#L607-L651

bin/utils.py
```diff
@@ -0,0 +1,49 @@
...
+    path_files_list = [directory + f for f in file_names_list]
```
mortendahl: If these are just strings you should use python's os.path.join instead: https://docs.python.org/3/library/os.path.html#os.path.join

bin/run
```diff
@@ -32,11 +33,21 @@ parser.add_argument(
...
+    '--data_dir', type=str, default="", help='Data directory with TFRecords file only')
 parser.add_argument(
     '--plaintext', action='store_true', help='Run the model in plaintext')
 parser.add_argument(
-    '--epochs', type=int, default=10, help='Number of times to run a prediction')
+    '--iterations', type=int, default=10, help='Number of times to run a prediction')
+parser.add_argument(
+    '--data_prep_path', type=str, default='', help='Path to a saved model with data pre-processing')
+parser.add_argument(
+    '--data_prep_start_node', type=str, default='serialized_example:0', help='Select data pre-processing start' \
```
jvmancuso: Should this have a default? I’m legitimately not sure

bin/run
```diff
@@ -32,11 +33,21 @@ parser.add_argument(
...
 parser.add_argument(
-    '--input_file', type=str, default="", help='load the data from a numpy file format (.npy)')
+    '--input_file', type=str, default="", help='Load the data from a numpy file format (.npy)')
+parser.add_argument(
+    '--data_dir', type=str, default="", help='Data directory with TFRecords file only')
 parser.add_argument(
     '--plaintext', action='store_true', help='Run the model in plaintext')
 parser.add_argument(
-    '--epochs', type=int, default=10, help='Number of times to run a prediction')
+    '--iterations', type=int, default=10, help='Number of times to run a prediction')
```
jvmancuso: +1 for this!

bin/run
jvmancuso: Agreed — this whole loop can be made a list comprehension that way

bin/run
jvmancuso: +1

bin/run
```diff
@@ -125,59 +150,87 @@ if config.plaintext:
...
-    pls = []
-    for i in range(len(input_spec)):
-        if i == 0:
-            name = "private-input/api/{}:0".format(i)
-        else:
-            name = "private-input_{}/api/{}:0".format(i, i)
-        pls.append(
-            tf.get_default_graph().get_tensor_by_name(name)
-        )
+    if config.data_prep_path:
```
jvmancuso: This if branch is identical to what comes later. Consider 
‘’‘
if not config.data_prep_path:
    # handle placeholders logic from lines 190-199
with tfe.Session():
    # etc.
‘’’

bin/utils.py
```diff
@@ -0,0 +1,49 @@
+import tensorflow as tf
```
jvmancuso: This filename should be more specific

bin/run
jvmancuso: I'd imagine as long as `tf.import_graph_def` happens within the device context when supplying this to `define_private_input`, that should be sufficient.  Otherwise, by passing the GraphDef directly we can iterate through it and manually override device NodeAttrs

yanndupis: makes sense. Is it ok if we  run on `master` this to load `gdef`?

```
with tf.Session(graph=tf.Graph()) as sess:
        tf.saved_model.loader.load(sess, ['serve'],  config.data_prep_path)
    
gdef = sess.graph_def
```

then pass that to `data_pred_from_saved_model`. I have validated that the ops get placed correctly on the `input_provider`.

bin/run
yanndupis: Definitely. let's change this.

bin/run
jvmancuso: Hm, perhaps we want to support both. Something like
```def data_prep_from_saved_model(saved_model_dir=None,  graph_def=None,...):
    if saved_model_dir is not None:
        # import on input provider
    if graph_def is not None:
        ...```

Comment (mortendahl) [reply bin/run]: @yanndupis if placement ends up being correct then I think it's alright to run that on master. Since you're loading into a new graph I wouldn't be surprised if the loaded savedmodel doesn't have any placements at all.

File: bin/run (hunk 4 above)
Comment (mortendahl) [reply bin/run]: We use it elsewhere if that helps

File: bin/run (hunk 2 above)
Comment (yanndupis) [reply bin/run]: Ok perfect, I will move forward with this. thanks.

File: bin/run (hunk 7 above)
Comment (yanndupis) [reply bin/run]: good point. this default are more specific to t2t. It's just a bit annoying to specify it every time we run a quick experiment locally. But i could remove it. It makes more sense in general.

File: bin/run (hunk 2 above)
Comment (jvmancuso) [reply bin/run]: I was thinking that this preprocessing might be sensitive information, as we've considered in the past.  If we want to support that situation early, it would be best to support both loading the graph on master, and delegating it to another party so that it loads with the proper device placements (which seems to be the case based on the Tensorboard photos @yanndupis has shared).

Comment (yanndupis) [reply bin/run]: Most of the comments should be addressed. Now I am loading the saved_model with data prep in master however the data pre-processing is happening on the `input_device`. 

@jvmancuso, I think this a good point. It's just I don't know how to avoid having tf.Session in `data_prep_from_saved_model` if we want to delegate the load step to the `input_device`.  

btw, the protocol buffer file with the model architecture and the weights are loaded from the master. so it's a similar situation.

**Comment by yanndupis** [comment]
Hey @mortendahl, @jvmancuso  

Thank you very much for your review. The comments should be addressed now. I have left a comment regarding the saved_model import. I am happy to change it if needed.

Thanks

File: bin/run (hunk 2 above)
Comment (mortendahl) [reply bin/run]: fyi both of the two options we've seen here loads the preprocessing step on master; we currently only support in-graph distribution. this bothered me in the previous code because it hinted that loading might happen on a specific device. one way to remember this is that *all* python code currently executes on master, no exceptions.

File: bin/run [hunk 11]
```diff
@@ -81,103 +97,121 @@ with gfile.GFile(model_file, 'rb') as f:
...
+
+    # SavedModel with data pre-processing steps
+    with tf.Session(graph=tf.Graph()) as sess:
+        tf.saved_model.loader.load(sess, ['serve'], config.data_prep_path)
+    
+    # Extract graph definition
+    data_prep_gdef = sess.graph_def
+
+    def provide_input():
+            return data_prep_from_saved_model(data_prep_gdef ,
```
mortendahl: there's an extra indent here

tf_encrypted/protocol/pond.py
```diff
@@ -907,6 +907,12 @@ def strided_slice(self, x: "PondTensor", *args: Any, **kwargs: Any):
 
         return x_sliced
 
+    @memoize
+    def gather(
```
mortendahl: how does this relate to https://github.com/tf-encrypted/tf-encrypted/pull/442?

yanndupis: oh no shouldn't be there. Let me fix this.

bin/run
jvmancuso: ah, that's unfortunate. getting the session outside of the provide_input fn makes the most sense

bin/run
jvmancuso: let's keep it in for now

bin/run
```diff
@@ -81,103 +97,121 @@ with gfile.GFile(model_file, 'rb') as f:
...
+
+if config.data_prep_path == '':
+    if config.input_file != '':
+        # TODO: Currently this allows for only one input
+        # If we want to load models with multiple different inputs
+        # We will need to save our inputs as a list of np.array
+        input_data = [np.load(config.input_file)]
+        assert len(input_data) == len(input_spec), 'the input_data length ({}) should be equal the number of inputs required by the loaded model ({})'.format(
+            len(input_data), len(input_spec))
+    else:
```
jvmancuso: There is an extra indent in this code block now

bin/run
```diff
@@ -81,103 +97,121 @@ with gfile.GFile(model_file, 'rb') as f:
...
+        input_data = []
+        for spec in input_spec:
```
jvmancuso: this can be done more succinctly in a list comprehension

yanndupis: Fixed

bin/run
jvmancuso: Ah my mistake!

yanndupis: no problem

yanndupis:
Hey guys, this PR should be ready. Thanks!

mortendahl:
good to go from my side when merge conflict is solved, well done @yanndupis! do you want to simply squash the commits or more do a more specific rewrite of history?

yanndupis:
@mortendahl , thanks! if you are ok, I will squash the commits. I am basically just adding a new feature and not changing the logic anywhere else.

mortendahl:
@yanndupis definitely good to go from my side!

PR #447, opened 2019-04-24
Release 0.5.2

mortendahl:
Release plus smaller changes required to make it go through

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -26,7 +26,7 @@ author = 'The tf-encrypted Contributors'
 # The short X.Y version
 version = '0.5.2'
 # The full version, including alpha/beta/rc tags
-release = '0.5.2-rc2'
+release = '0.5.2'
 
 
 # -- General configuration ---------------------------------------------------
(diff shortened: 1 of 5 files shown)


mortendahl:
TFE tests run on rc2

docs/source/conf.py
```diff
@@ -24,9 +24,9 @@
 author = 'The tf-encrypted Contributors'
 
 # The short X.Y version
-version = '0.5.1'
+version = '0.5.2'
 # The full version, including alpha/beta/rc tags
-release = '0.5.1'
+release = '0.5.2-rc2'
```
jvmancuso: if this is the commit for the 0.5.2 release, I believe the rc tags need to be removed in each of these files -- hopefully @ianlivingstone can confirm here

ianlivingstone: Yep! They need to be updated for each release. Ideally, these values should be centralized in one file or templated and derived at the time of the release so no one has to think about it.

mortendahl: kept them in in case we needed a rc3 -- if none of you think so i'll prepare the final release

PR #451, opened 2019-04-25
Examples from O'Reilly AI talk

mortendahl:


examples/logistic/common.py
```diff
@@ -0,0 +1,159 @@
...
+        self.b = tfe.define_private_variable(tf.zeros([1]))
+        self.b_masked = tfe.mask(self.b)
+
+    @property
+    def weights(self):
+        return self.w, self.b
+
+    def forward(self, x):
+        with tf.name_scope("forward"):
+            out = tfe.matmul(x, self.w_masked) + self.b_masked
```
jvmancuso: Would it be worth using tfe.layers.Dense here?

jvmancuso: (although perhaps best to wait until the layer implementations have been cleaned up)

examples/logistic/common.py
```diff
@@ -0,0 +1,159 @@
...
+    @property
+    def initializer(self):
+        return tf.group(self.train_initializer, self.test_initializer)
+
+    def provide_training_data(self):
+
+        def norm(x, y):
+            return tf.cast(x, tf.float32), tf.expand_dims(y, 0)
+
+        x_raw = tf.random.uniform(
```
jvmancuso: how much work would it be to replace this with real data? if not this PR, maybe open an issue for it

examples/logistic/common.py
mortendahl: definitely, even thought the model could eventually move into TFE. suggest we do that as part of high-level API update!

examples/logistic/common.py
mortendahl: all for this if/when we find the right use case!

PR #481, opened 2019-05-15
Keras Dense Layer

yanndupis:
Hey guys, 

Here is an implementation of the the keras dense layer. I didn't include anything related to backprop as we are focusing on private prediction for the initial implementation. 

Thanks


tf_encrypted/keras/engine/base_layer.py
```diff
@@ -0,0 +1,93 @@
...
+      self.build(input_shapes)
+
+      self.built = True
+
+    outputs = self.call(inputs, *args, **kargs)
+
+    return outputs
+
+  @property
+  def prot(self) -> Optional[Protocol]:
```
mortendahl: I'd suggest using `protocol` instead of `prot`, since everything else seems to use a somewhat explicit naming scheme (which I like)

tf_encrypted/keras/layers/dense.py
```diff
@@ -0,0 +1,121 @@
...
+import numpy as np
+import tensorflow as tf
+from tensorflow.python.keras import initializers
+
+from tf_encrypted.keras.engine.base_layer import Layer
+from tf_encrypted.keras import activations
+from tf_encrypted.protocol.pond import PondPublicTensor, PondPrivateTensor
+
+
+InitialTensor = Optional[Union[np.ndarray,
```
mortendahl: this is not used anywhere

tf_encrypted/keras/layers/dense.py
```diff
@@ -0,0 +1,121 @@
+# pylint: disable=arguments-differ
+"""Dense (i.e. fully connected) Layer implementation."""
+from typing import Union, Optional
+import numpy as np
+import tensorflow as tf
+from tensorflow.python.keras import initializers
+
+from tf_encrypted.keras.engine.base_layer import Layer
```
mortendahl: the less specific `from tf_encrypted.keras.engine import Layer` should work as well

tf_encrypted/keras/layers/dense.py
```diff
@@ -0,0 +1,121 @@
...
+from tf_encrypted.keras import activations
+from tf_encrypted.protocol.pond import PondPublicTensor, PondPrivateTensor
```
mortendahl: this is bad. seems the only purpose is typing, in which case we should make sure that typing is really needed

tf_encrypted/keras/layers/dense.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+InitialTensor = Optional[Union[np.ndarray,
+                               tf.Tensor, PondPublicTensor, PondPrivateTensor]]
+
+
+argument_msg = "`{}` argument is not implemented for layer {}"
```
mortendahl: `arg_not_impl_msg`?

tf_encrypted/keras/layers/dense.py
```diff
@@ -0,0 +1,121 @@
...
+                                                    "Dense"))
+    if kernel_constraint:
+      raise NotImplementedError(argument_msg.format("kernel_constraint",
+                                                    "Dense"))
+    if bias_constraint:
+      raise NotImplementedError(argument_msg.format(" bias_constraint",
+                                                    "Dense"))
+
+  def compute_output_shape(self, input_shape):
+    return [input_shape[0] + self.units]
```
mortendahl: is this consistent with https://github.com/tf-encrypted/tf-encrypted/pull/481/files#diff-a4f929bc8ab0de50f08cd8542bd67801R46 and https://github.com/tf-encrypted/tf-encrypted/pull/481/files#diff-a4f929bc8ab0de50f08cd8542bd67801R51 ?

tf_encrypted/keras/layers/dense.py
```diff
@@ -0,0 +1,121 @@
...
+
+  def build(self, input_shape) -> None:
+    units_in = int(input_shape[1])
```
mortendahl: same question as in `compute_output_shape`

tf_encrypted/keras/activations.py
```diff
@@ -0,0 +1,39 @@
+"""Provide activation functions"""
+from tf_encrypted.protocol.protocol import get_protocol
```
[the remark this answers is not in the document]
mortendahl: Once the APIs settle a bit it will probably be best to *not* use relative imports but do what you're doing here instead -- creates better modularity. So perhaps stay away from relative imports here 👍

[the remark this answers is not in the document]
jvmancuso: I like this one `from tf_encrypted import get_protocol`

tf_encrypted/keras/engine/base_layer.py
```diff
@@ -0,0 +1,93 @@
...
+  * `build()`: Called once from `__call__`, when we know the shapes of inputs
+    and `dtype`.
+  * `call()`: Called in `__call__` after making sure `build()` has been called
+    once. Should actually perform the logic of applying the layer to the
+    input tensors (which should be passed in as the first argument).
+  """
+
+  def __init__(self, trainable=True, **kwargs):
+
+    allowed_kwargs = {
```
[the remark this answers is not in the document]
jvmancuso: a few of these will be needed when we get to `Sequential` -- at least `input_shape` is necessary in the first layer of a `Sequential` model container

tf_encrypted/keras/engine/base_layer.py
```diff
@@ -0,0 +1,93 @@
...
+    """This is where the layer's logic lives.
+    Arguments:
+        inputs: Input tensor, or list/tuple of input tensors.
+    Returns:
+        A tensor or list/tuple of tensors.
+    """
+    return inputs
+
+  @abstractmethod
+  def compute_output_shape(self, input_shape) -> List[int]:
```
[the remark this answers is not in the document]
jvmancuso: agreed, perhaps it's best to leave this type out as well (since the notion of a shape is pretty well understood)

tf_encrypted/keras/engine/base_layer.py
```diff
@@ -0,0 +1,93 @@
...
+    # Validate optional keyword arguments.
+    for kwarg in kwargs:
+      if kwarg not in allowed_kwargs:
+        raise TypeError('Keyword argument not understood:', kwarg)
+
+    self.trainable = trainable
+    self.built = False
+
+  @abstractmethod
+  def build(self, input_shape) -> None:
```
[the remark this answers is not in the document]
yanndupis: Agreed, what do you guys think about not having types in the keras API for now? I don't think it's adding much value in this context.

tf_encrypted/keras/engine/base_layer.py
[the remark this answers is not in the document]
yanndupis: Exactly, it's more like a placeholder for now but we are going to need it when we work on `Sequential`.

tf_encrypted/keras/engine/base_layer.py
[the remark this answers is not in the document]
mortendahl: Yeah works for me 👍

tf_encrypted/keras/engine/base_layer.py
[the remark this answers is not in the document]
mortendahl: should the verification logic be there then instead?

tf_encrypted/keras/activations.py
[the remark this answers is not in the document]
yanndupis: good point.

tf_encrypted/keras/engine/base_layer.py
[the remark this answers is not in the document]
yanndupis: I think it should be ok. Every layer could expect one of these arguments. That's how it's currently done in the keras package.

tf_encrypted/keras/layers/dense.py
yanndupis: I think it is. But it needs a coma instead of a + sign.

tf_encrypted/keras/engine/base_layer.py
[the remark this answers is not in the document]
mortendahl: alright 👍

tf_encrypted/keras/layers/dense.py
mortendahl: does it not limit the output shape to two dimension?

yanndupis: ah yes, you are right. I guess we could update the comment to 2D tensor (as an input and output) for the Dense. No sure we want to go beyond 2d tensors for now for the dense layer.

jvmancuso: It does, and it's on purpose.  Dense layers should only output rank 2 tensors according to the Keras spec.  Gemm is not implemented in Keras afaik

yanndupis: @jvmancuso do you have an opinion regarding this?

mortendahl: From [here](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Dense):

> Note: if the input to the layer has a rank greater than 2, then it is flattened prior to the initial dot product with kernel.

Is it worth adding that logic?

jvmancuso: I normally do the Flatten explicitly

jvmancuso: but if that's how they do it we should aim for the same

yanndupis: Same here. And in the official Keras examples, they also do Flatten explicitly:
https://github.com/keras-team/keras/blob/master/examples/mnist_cnn.py

tf_encrypted/keras/engine/base_layer.py
```diff
@@ -1,10 +1,7 @@
 """Includes base classes used by all layer types."""
 
-from typing import List, Optional
 from abc import ABC, abstractmethod
-
-from tf_encrypted.protocol.pond import TFEVariable
-from tf_encrypted.protocol.protocol import get_protocol, Protocol
+from tf_encrypted import get_protocol
```
mortendahl: consider adding blank link here to separate standard packages and project packages

yanndupis:
Hey guys, thank you so much for your review. I think all the comments should be addressed now.

tf_encrypted/keras/layers/dense.py
mortendahl: there's a bit more going on: https://github.com/tensorflow/tensorflow/blob/r1.13/tensorflow/python/keras/layers/core.py#L966-L973

if you guys don't think people will fall into to (including when they load models) then I'm okay to skip

tf_encrypted/keras/engine/base_layer.py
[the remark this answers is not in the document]
mortendahl: there's a `None` type here fyi

tf_encrypted/keras/layers/dense.py
yanndupis: ah, I missed this comment.

jvmancuso: I think if tf.keras does it we should follow suit.  it should be an easy addition.

jvmancuso:
I think we should address the last few comments before merging but also think it's nearly there!

yanndupis:
Hey guys, I have added flatten when the input rank is greater than 2, removed `None` etc. Let me know what you think.

tf_encrypted/keras/layers/dense.py
```diff
@@ -0,0 +1,118 @@
...
+
+  def call(self, inputs):
+
+    input_shape = inputs.shape.as_list()
+
+    #  if the input to the layer has a rank greater than 2,
+    #  then it is flattened prior to the initial dot product with kernel.
+    if len(input_shape) > 2:
+      flat_shape = [input_shape[0]] + [np.prod(input_shape[1:])]
+      inputs = inputs.reshape(flat_shape)
```
mortendahl: you should be able to simply use `flat_shape = [input_shape[0], -1]` here, avoiding the numpy import

yanndupis: ah yes! sorry about that.

tf_encrypted/keras/layers/dense.py
jvmancuso: this is going to break if the input_shape has rank greater than 2

jvmancuso: I'd also recommend an assert here that checks to make sure units is equal to dimension 1 of the flattened input tensor

jvmancuso: and this means you can store the flattened input_shape as an attribute for use in `call`, so that can be simplified a bit

yanndupis: ah good catch

tf_encrypted/keras/layers/dense.py
```diff
@@ -83,6 +83,14 @@ def compute_output_shape(self, input_shape):
     return [input_shape[0], self.units]
 
   def build(self, input_shape):
+
+    rank = len(input_shape)
+
+    if rank > 2:
+      raise Exception(
```
mortendahl: maybe use `NotImplementedError` instead

yanndupis: correct, commit arriving in 1 sec

PR #486, opened 2019-05-17
Add Activation layer

yanndupis:
Hey, 

This PR add the [Activation layer](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Activation). 

Thanks,

### DIFF
--- a/tf_encrypted/keras/layers/activation_layer.py
+++ b/tf_encrypted/keras/layers/activation_layer.py
@@ -5,7 +5,7 @@ from tf_encrypted.keras import activations
 class Activation(Layer):
   """Applies an activation function to an output.
   Arguments:
-      activation: name of activation function to use
+      activation: name of activation function to use or
           TF Encrypted operation.
   Input shape:
       Arbitrary. Use the keyword argument `input_shape`
--- a/tf_encrypted/keras/layers/activation_layer.py
+++ b/tf_encrypted/keras/layers/activation_layer.py
@@ -1,0 +1,29 @@
+"""Activation Layer implementation."""
+from tf_encrypted.keras.engine.base_layer import Layer
+from tf_encrypted.keras import activations
+
+class Activation(Layer):
+  """Applies an activation function to an output.
+  Arguments:
+      activation: name of activation function to use
+          TF Encrypted operation.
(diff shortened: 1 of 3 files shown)


tf_encrypted/keras/layers/activation_layer.py
```diff
@@ -0,0 +1,29 @@
...
+    self.activation = activations.get(activation)
+
+  def build(self, input_shape):
+    pass
+
+  def call(self, inputs):
+    return self.activation(inputs)
+
+  def compute_output_shape(self, input_shape):
+    return input_shape
```
mortendahl: ideally this should be forwarded to the activation function in case it changes the shape. however, this adds some complexity so maybe it's best simply to add a comment in the docstring along the lines of "These are assumed to preserve shape."

tf_encrypted/keras/layers/activation_layer_test.py
```diff
@@ -0,0 +1,52 @@
...
+
+        out_pond = sess.run(out.reveal())
+
+    #reset graph
+    tf.reset_default_graph()
+
+    with tf.Session() as sess:
+      x_in = tf.Variable(x, dtype=tf.float32)
+
+      activation_tf = tf.keras.layers.Activation(**layer_kwargs)
```
mortendahl: we are getting very close to the point where you can re-use code like this, and in one case pass in `tf` to the core test, and in the other `tfe`! (you could do it here for instance)

tf_encrypted/keras/layers/activation_layer.py
jvmancuso: Do you have any in mind?  I don't know of any activation functions that change the shape of the inputs.

tf_encrypted/keras/layers/activation_layer_test.py
jvmancuso: this calls for some test automation like in test_convert.py :)

tf_encrypted/keras/layers/activation_layer.py
mortendahl: oh no, simply saying that there'll be a bug if that ever happens

PR #488, opened 2019-05-20
Add additional normalization, checks to Conv2D attributes

yanndupis:
Hey, 

This PR adds some additional normalization, checks to Conv2D attributes. 

Thank you,


tf_encrypted/keras/layers/convolutional.py
```diff
@@ -4,9 +4,11 @@
 
 import numpy as np
 from tensorflow.python.keras import initializers
+from tensorflow.python.keras.utils import conv_utils
 
 from tf_encrypted.keras.engine import Layer
 from tf_encrypted.keras import activations
+from . import layers_utils
```
jvmancuso: please stay with absolute imports since the file is already using those.

tf_encrypted/keras/layers/convolutional.py
```diff
@@ -167,7 +175,10 @@ def call(self, inputs):
     if self.data_format != 'channels_first':
       inputs = self.prot.transpose(inputs, perm=[0, 3, 1, 2])
 
-    outputs = self.prot.conv2d(inputs, self.kernel, self.strides, self.padding)
+    outputs = self.prot.conv2d(inputs,
+                               self.kernel,
+                               self.strides[0],
```
jvmancuso: eventually we will want to fix this in our own implementation, but okay for now

tf_encrypted/keras/layers/layers_utils.py
```diff
@@ -0,0 +1,13 @@
+# pylint: disable=arguments-differ
```
jvmancuso: why is this here?

jvmancuso:
Also, can we add some more tests for this layer? there is a lot of functionality in the constructor that's not being tested with the current set

yanndupis:
Thanks for your review @jvmancuso, I will add more tests.

tf_encrypted/keras/layers/layers_utils.py
yanndupis: shouldn't be

PR #494, opened 2019-05-21
Add Flatten layer

yanndupis:
Hey, 

This PR add [Flatten](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Flatten). I have left the data_format attribute. However it doesn't do anything under the hood as it doesn't seem to add value for our use. Should we add a log message for that? I don't think it should return `NotImplementedError` in this context.

Thanks

### DIFF
--- a/tf_encrypted/keras/layers/flatten.py
+++ b/tf_encrypted/keras/layers/flatten.py
@@ -49,7 +49,7 @@ class Flatten(Layer):
 
   def compute_output_shape(self, input_shape):
     if not input_shape:
-      output_shape = [1]
+      raise ValueError("input_shape shouldn't be empty or None")
     output_shape = [input_shape[0]]
     if all(input_shape[1:]):
       output_shape += [np.prod(input_shape[1:])]
--- a/tf_encrypted/keras/layers/flatten.py
+++ b/tf_encrypted/keras/layers/flatten.py
@@ -30,8 +30,15 @@ class Flatten(Layer):
 
   def call(self, inputs):
     input_shape = inputs.shape.as_list()
+    rank = len(input_shape)
 
-    if len(input_shape) == 1:
+    if self.data_format == 'channels_first' and rank > 1:
+      permutation = [0]
+      permutation.extend([i for i in range(2, rank)])
+      permutation.append(1)
+      inputs = self.prot.transpose(inputs, perm=permutation)
+
+    if rank == 1:
       flatten_shape = [input_shape[0], 1]
     else:
       flatten_shape = [input_shape[0], -1]
--- a/tf_encrypted/keras/layers/flatten.py
+++ b/tf_encrypted/keras/layers/flatten.py
@@ -1,0 +1,52 @@
+"""Flatten Layer implementation."""
+
+import numpy as np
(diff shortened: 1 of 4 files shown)


tf_encrypted/keras/layers/flatten.py
```diff
@@ -0,0 +1,52 @@
...
+  """
+
+  def __init__(self, data_format=None, **kwargs):
+    super(Flatten, self).__init__(**kwargs)
+    self.data_format = conv_utils.normalize_data_format(data_format)
+
+  def build(self, input_shape):
+    self.built = True
+
+  def call(self, inputs):
```
jvmancuso: I suggest checking how data_format is used for this layer in tf.keras [here](https://github.com/tensorflow/tensorflow/blob/1511803cb3f7f1fe88970c406abfe8f2c07e68a0/tensorflow/python/keras/layers/core.py#L571) and following suit

tf_encrypted/keras/layers/flatten.py
```diff
@@ -0,0 +1,52 @@
...
+    if len(input_shape) == 1:
+      flatten_shape = [input_shape[0], 1]
+    else:
+      flatten_shape = [input_shape[0], -1]
+
+    outputs = self.prot.reshape(inputs, flatten_shape)
+
+    return outputs
+
+  def compute_output_shape(self, input_shape):
```
jvmancuso: the code in this function is broken -- the code in the if block is unreachable if input_shape is always well-defined (i.e. non-empty and not None), but the code below the if block will fail with an IndexError if input_shape is allowed to be either empty and a TypeError if input_shape is allowed to be None.

suggest you revisit the TensorShape class in TF and see if it's useful.  if so, we should discuss with @mortendahl if we should be using it more often throughout the codebase, and if not, you should figure out what the intended behavior is and simulate it more precisely with plain Python

tf_encrypted/keras/layers/flatten.py
yanndupis: Sounds good. Just added a commit. It's now supporting channels_first.

tf_encrypted/keras/layers/flatten.py
yanndupis: Thanks good catch. I am now raising an error when input_shape is empty or None. 

@mortendahl, @jvmancuso and I had a discussion about TensorShape. it comes with several functionalities when dealing with shapes. I will open an issue so we can discuss if we want to use it across the code base. Thanks

yanndupis:
I think it's ready to be merged. Thanks

tf_encrypted/keras/layers/flatten.py
mortendahl: Sounds good 👍 one problem with partial shapes is the triple stuff, since they can't be generated offline if the shape is not complete. We could make that a special case though, as the online triple source should work with partial shapes (although we may have to add some control dependencies).

jvmancuso:
on a roll!

tf_encrypted/keras/layers/flatten.py
yanndupis: Sounds good, that totally makes sense.

PR #501, opened 2019-05-23
minor bin script fixes

justin1121:
Also add some more files to the .dockerignore file.


.dockerignore
```diff
@@ -1,2 +1,23 @@
...
+.mypy_cache/
+.pytest_cache/
+
+docs/build
+dist/
+tmp/
+**/*.pb
+data/
+*.egg-info
+*.so
```
mortendahl: some repetition here

bin/serve
```diff
@@ -89,13 +90,13 @@ tfe.set_tfe_trace_flag(enable_traces)
...
 
 remote_config = tfe.RemoteConfig(
-    {
-        'master': master_address,
-        'server0': server0_address,
-        'server1': server1_address,
-        'crypto-producer': producer_address,
-        'weights-provider': provider_address
-    }
+    OrderedDict([
```
mortendahl: 👍 you could probably also use 
```python
OrderedDict(
  master=...,
  server0=..., )
```

justin1121: Much nicer! Thanks.

justin1121: although its doesn't like names like `crypto-producer` in the argument list. Will leave it how it is for now.

justin1121: Doens't like the `-` character.

PR #507, opened 2019-05-27
channels_last support to batchnorm layer

avitalsh:
Added channels_last (NHWC format) support to batch norm layer and updated the unitest accordingly. 
Channels_first remains the default.

### DIFF
--- a/tests/test_batchnorm.py
+++ b/tests/test_batchnorm.py
@@ -14,49 +14,85 @@ class TestBatchnorm(unittest.TestCase):
 
   def test_forward(self) -> None:
 
-    batch_size, channels_in, img_height, img_width = (32, 3, 28, 28)
+    def run_test(input_batchnorm, input_shape, mean, variance, scale, offset, variance_epsilon, channels_first):
 
-    input_shape = [batch_size, channels_in, img_height, img_width]
-    input_batchnorm = np.random.normal(size=input_shape).astype(np.float32)
+      with tfe.protocol.Pond() as prot:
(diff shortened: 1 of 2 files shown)


tests/test_batchnorm.py
```diff
@@ -14,50 +14,86 @@ def setUp(self):
...
+          sess.run(tf.global_variables_initializer())
 
-        out_tensorflow = sess.run(batchnorm_out_tf)
+          out_tensorflow = sess.run(batchnorm_out_tf)
 
-        np.testing.assert_array_almost_equal(
+          np.testing.assert_array_almost_equal(
             out_pond, out_tensorflow, decimal=1)
 
+    def test_channels_first():
```
jvmancuso: can you move these two out into their own methods on TestBatchnorm? that way if one of them fails in the future it's a bit more informative for the person running this file

jvmancuso: you can also look into using `pytest.mark.parametrize` for this if you're feeling ambitious! (but not necessary for merging)

avitalsh: @jvmancuso Thanks for the review. I separated the tests. Tried looking at ```pytest.mark.parametrize``` but it seems to be not working great with ```unittest.TestCase``` subclasses -
https://stackoverflow.com/questions/35561843/py-test-parametrize-test-cases-from-classes
https://stackoverflow.com/questions/18182251/does-pytest-parametrized-test-work-with-unittest-class-based-tests

jvmancuso: I see, in that case parameterization is probably beyond the scope of this PR. LGTM!

PR #509, opened 2019-05-28
Basic Sequential model class

jvmancuso:
Provides a Sequential class that implements a subset of functionality of the original TF Keras version


tf_encrypted/keras/layers/dense_test.py
```diff
@@ -37,6 +37,9 @@ def _core_dense(self, **layer_kwargs):
     agreement_test(tfe.keras.layers.Dense,
                    kwargs=kwargs,
                    input_shape=input_shape)
+    layer_test(tfe.keras.layers.Dense,
```
yanndupis: Do we want this test only for `Dense`, `Conv2D` and `Activation` or for all the layers to be consistent?

tf_encrypted/keras/engine/sequential.py
```diff
@@ -0,0 +1,104 @@
...
+
+    elif self.outputs:
+      # If the model is being built continuously on top of an input layer:
+      # refresh its output.
+      output_tensor = layer(self.outputs[0])
+      if isinstance(output_tensor, list):
+        raise TypeError('All layers in a Sequential model '
+                        'should have a single output tensor. '
+                        'For multi-output layers, '
+                        'use the functional API.')
```
mortendahl: do we have this yet? or is it part of not removing too much since we'll do this in the near future?

tf_encrypted/keras/engine/base_layer.py
```diff
@@ -88,3 +100,15 @@ def __call__(self, inputs, *args, **kargs):
   @property
   def prot(self):
     return get_protocol()
+
+  @property
+  def name(self):
+    return self._name
+
+  def _init_set_name(self, name, zero_based=True):
```
mortendahl: is this normal behaviour? wouldn't users expect layers to have a name close to what they specified? (ie perhaps with an added counter but that's it)

tf_encrypted/keras/__init__.py
```diff
@@ -1,12 +1,13 @@
 """Higher-level layer abstractions built on TF Encrypted."""
 from __future__ import absolute_import
 
-from . import engine
-from . import layers
-
+from tf_encrypted.keras import engine
+from tf_encrypted.keras import layers
+from tf_encrypted.keras.engine.sequential import Sequential
```
mortendahl: `Sequential` should be in `tfe.keras.models` instead of `tfe.keras.engine` per both TF and Keras

jvmancuso: this is not the case: https://github.com/tensorflow/tensorflow/blob/r1.13/tensorflow/python/keras/engine/sequential.py
it just happens to be publicly available from that location

tf_encrypted/keras/engine/base_layer.py
jvmancuso: this is normal behavior (copied nearly verbatim from the keras version of this class or a superclass). the default name is to take the class name and snake case it, but if a user specifies their own name to the constructor it will be used instead (see where this gets called in the constructor for specifics)

tf_encrypted/keras/engine/sequential.py
```diff
@@ -0,0 +1,104 @@
+"""Sequential model API."""
+import tf_encrypted as tfe
+from tf_encrypted.keras.engine.base_layer import Layer
+from tf_encrypted.keras.engine.input_layer import InputLayer, Input
+
+
+class Sequential(Layer):
+  """Model defined by a stack of layers in sequence.
+
+  TODO
```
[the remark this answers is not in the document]
jvmancuso: I say we leave it and fix later

tf_encrypted/keras/engine/sequential.py
jvmancuso: yup, part of not removing too much

tf_encrypted/keras/layers/dense_test.py
jvmancuso: probably all the layers

tf_encrypted/keras/__init__.py
mortendahl: public API location is what matters, not the file it lives in

PR #514, opened 2019-05-29
fix with protocol

justin1121:
Closes https://github.com/tf-encrypted/tf-encrypted/issues/439

Just need to use the global `set_protocol` instead of the one inside the `protocol.py` file. Hope this makes sense!

### DIFF
--- a/tf_encrypted/__init__.py
+++ b/tf_encrypted/__init__.py
@@ -46,6 +46,7 @@ def set_protocol(prot: Optional[protocol.Protocol] = None) -> None:
       globals()[name] = func
 
   # record new protocol
+  prot.clear_initializers()
   protocol.set_protocol(prot)
 
 
--- a/tf_encrypted/protocol/protocol.py
+++ b/tf_encrypted/protocol/protocol.py
@@ -6,6 +6,7 @@ from types import TracebackType
 
 import tensorflow as tf
 
+import tf_encrypted as tfe
 from ..tensor.factory import AbstractTensor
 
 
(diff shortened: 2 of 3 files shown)


tf_encrypted/__init__.py
```diff
@@ -46,6 +46,7 @@ def set_protocol(prot: Optional[protocol.Protocol] = None) -> None:
       globals()[name] = func
 
   # record new protocol
+  prot.clear_initializers()
```
justin1121: Run this function every time the protocol changes created a weird side effect where the tests would fail because of ops not belonging to the same graph. This is caused `inspect.getmembers(prot, predicate=inspect.ismethod)` function call. When it iterates through all the members of the protocol it access `.initializers` which is a property. Accessing it like this actually makes it so it gets called. Seems like weird behaviour to me. Clearing the initializers here fixes this issue but I'm not sure it's actually what we want. I feel like the real fix is the `.initializers` property not getting called but not sure there is anything we do about that.

Ideas are welcome.

mortendahl: yeah I don't think this is want we want, since we might loose initializers this way.

tf_encrypted/protocol/protocol.py
```diff
@@ -6,6 +6,7 @@
 
 import tensorflow as tf
 
+import tf_encrypted as tfe
```
mortendahl: not a big fan of the mixing importing style. i think we should rewrite the way this works instead tbh.

justin1121: Yeah I'm not sure the best way to handle this.

tf_encrypted/__init__.py
justin1121: Think we'll need a different way of exposing the functions globally than what we're currently doing.

justin1121:
This isn't a good fix. Closing for now.

PR #523, opened 2019-05-30
First version of TFE daemon

mortendahl:


mortendahl:
Maybe we don't need a daemon, maybe we just need a better player

mortendahl:
Closing until unblocked

PR #524, opened 2019-05-30
Keras batchnorm

justin1121:
Closes https://github.com/tf-encrypted/tf-encrypted/issues/515

### DIFF
--- a/tf_encrypted/keras/layers/normalization_test.py
+++ b/tf_encrypted/keras/layers/normalization_test.py
@@ -33,7 +33,7 @@ class TestBatchNormalization(unittest.TestCase):
 
   def _core_batchnorm(self, input_shape, **layer_kwargs):
     base_kwargs = {
-      'fused': False
+        'fused': False
     }
 
     kwargs = {**base_kwargs, **layer_kwargs}
--- a/tf_encrypted/keras/layers/normalization_test.py
+++ b/tf_encrypted/keras/layers/normalization_test.py
@@ -35,8 +35,9 @@ class TestBatchNormalization(unittest.TestCase):
     const = np.random.normal(input_shape)
     initializer = tf.keras.initializers.Constant(const)
 
-    self._core_batchnorm([1, 1, 1, 3], moving_mean_initializer=initializer)
-    self._core_batchnorm([1] + input_shape, moving_variance_initializer=initializer)
+    self._core_batchnorm([1] + input_shape, moving_mean_initializer=initializer)
+    self._core_batchnorm([1] + input_shape,
+                         moving_variance_initializer=initializer)
 
   def _core_batchnorm(self, input_shape, **layer_kwargs):
     base_kwargs = {
--- a/tf_encrypted/keras/layers/normalization_test.py
+++ b/tf_encrypted/keras/layers/normalization_test.py
@@ -30,12 +30,18 @@ class TestBatchNormalization(unittest.TestCase):
   def test_batchnorm_no_center(self):
     self._core_batchnorm([1, 1, 1, 3], center=False)
 
-  def test_batchnorm_non_default_init(self):
+  def test_batchnorm_non_default_mean_init(self):
     input_shape = [1, 1, 3]
     const = np.random.normal(input_shape)
     initializer = tf.keras.initializers.Constant(const)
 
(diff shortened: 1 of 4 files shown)


tf_encrypted/keras/layers/normalization_test.py
```diff
@@ -0,0 +1,50 @@
...
+  def test_batchnorm_no_scale(self):
+    self._core_batchnorm([1, 1, 1, 3], scale=False)
+
+  def test_batchnorm_no_center(self):
+    self._core_batchnorm([1, 1, 1, 3], center=False)
+
+
+  def _core_batchnorm(self, input_shape, **layer_kwargs):
+    base_kwargs = {
+        'fused': False
```
yanndupis: Out of curiosity, what happen if fused = True?

justin1121: The tensorflow fused implementation is limited in where the channel can be. Can't remember if its first or last but the test will fail when it tries to compare the keras answer to our answer because it can't calculate the keras answer.

tf_encrypted/keras/layers/normalization_test.py
```diff
@@ -0,0 +1,50 @@
...
+class TestBatchNormalization(unittest.TestCase):
+
+  def setUp(self):
+    tf.reset_default_graph()
+
+  def test_batchnorm_2d(self):
+    self._core_batchnorm([1, 4], axis=1)
+
+  def test_batchnorm_4d(self):
+    self._core_batchnorm([1, 1, 1, 3])
```
yanndupis: I would suggest to have a test where `moving_mean_initializer` and `moving_variance_initializer` are not set to the defaults zeros and ones respectively.

tf_encrypted/keras/layers/normalization_test.py
yanndupis: Ok great, that makes sense.

avitalsh:
Hey @justin1121, I was comparing the tfe.layers.batchnorm implementation and the tfe.keras.layers bathnorm (because I want to add batchnorm to the converter) and I'm not sure but I think there is a small mistake.
If I understand correctly, the translation between the keras api and the tf.nn api means that -
```scale``` (tf.nn) = ```gamma``` (tf.keras) and ```offset``` (tf.nn) = ```beta``` (tf.keras).

In the nn batchnorm we have this two ```if``` statments:
https://github.com/tf-encrypted/tf-encrypted/blob/d48de7ab37e270fac2a0ae1fc6de87cb6bd9ce6c/tf_encrypted/layers/batchnorm.py#L76-L79

but in the keras batchnorm we have:
https://github.com/tf-encrypted/tf-encrypted/blob/d48de7ab37e270fac2a0ae1fc6de87cb6bd9ce6c/tf_encrypted/keras/layers/normalization.py#L202-L205

So I think this line
https://github.com/tf-encrypted/tf-encrypted/blob/d48de7ab37e270fac2a0ae1fc6de87cb6bd9ce6c/tf_encrypted/keras/layers/normalization.py#L202

should be replaced with this line
https://github.com/tf-encrypted/tf-encrypted/blob/d48de7ab37e270fac2a0ae1fc6de87cb6bd9ce6c/tf_encrypted/keras/layers/normalization.py#L204

**Comment by justin1121** [comment]
Hi @avitalsh, I'm sure I see what you are seeing. Can you give me a whole snippet (the call method) of what you think the code **should** look like. You could even submit a PR with the changes and we can evaluate there 😄

**Comment by avitalsh** [comment]
After looking at this again I think I was wrong - the keras batchnorm is fine but there is a mix up in the layers.batchnorm.

This is the code now - 
https://github.com/tf-encrypted/tf-encrypted/blob/d48de7ab37e270fac2a0ae1fc6de87cb6bd9ce6c/tf_encrypted/layers/batchnorm.py#L74-L81

and I think it should be like this:
```
    if self.scale is None and self.offset is None:
      out = (x - self.mean) * self.denom
    elif self.offset is None:
      out = self.scale * (x - self.mean) * self.denom
    elif self.scale is None:
      out = (x - self.mean) * self.denom + self.offset
    else:
      out = self.scale * (x - self.mean) * self.denom + self.offset
```

(switched lines 76 and 78)

I'll submit a PR later

Issue #499, opened 2019-05-22
Use TensorShape for shapes in TFE


yanndupis:
In the keras code base they use [tf.TensorShape](https://www.tensorflow.org/api_docs/python/tf/TensorShape) at different places, for example [here](https://github.com/tensorflow/tensorflow/blob/r1.13/tensorflow/python/keras/layers/core.py#L983). 

We should consider using [tf.TensorShape](https://www.tensorflow.org/api_docs/python/tf/TensorShape) as well, so we can handle properly partially-known shapes and unknown shapes.

jvmancuso:
from @mortendahl in #494 
>one problem with partial shapes is the triple stuff, since they can't be generated offline if the shape is not complete. We could make that a special case though, as the online triple source should work with partial shapes (although we may have to add some control dependencies).

jvmancuso:
yeah, currently the online triples don't work either (factories try to generate tensors with partially known shapes, which gives a ValueError since tensors need to have well-defined shapes)

PR #525, opened 2019-05-30
add keras reshape layer

justin1121:
Closes https://github.com/tf-encrypted/tf-encrypted/issues/516

### DIFF
--- a/tf_encrypted/keras/layers/core.py
+++ b/tf_encrypted/keras/layers/core.py
@@ -1,0 +1,93 @@
+"""Core layers such as Reshape"""
+
+from tf_encrypted.keras.engine import Layer
+
+class Reshape(Layer):
+  """Reshapes an output to a certain shape.
+  Arguments:
+    target_shape: Target shape. Tuple of integers,
+      does not include the samples dimension (batch size).
+  Input shape:
+    Arbitrary, although all dimensions in the input shaped must be fixed.
+    Use the keyword argument `input_shape`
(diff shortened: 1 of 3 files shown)


tf_encrypted/keras/layers/core.py
```diff
@@ -0,0 +1,93 @@
...
+      output_shape += self._fix_unknown_dimension(input_shape[1:],
+                                                  self.target_shape)
+    return output_shape
+
+  def call(self, inputs):
+    return inputs.reshape((int(inputs.shape[0]),) + self.target_shape)
+
+  def get_config(self):
+    config = {'target_shape': self.target_shape}
+    base_config = super(Reshape, self).get_config()
```
yanndupis: Unfortunately, we haven't implemented `get_config` in the base Layer class.

tf_encrypted/keras/layers/core.py
```diff
@@ -0,0 +1,93 @@
...
+    if unknown is not None:
+      if known == 0 or original % known != 0:
+        raise ValueError(msg)
+      output_shape[unknown] = original // known
+    elif original != known:
+      raise ValueError(msg)
+    return output_shape
+
+  def compute_output_shape(self, input_shape):
+    input_shape = tensor_shape.TensorShape(input_shape).as_list()
```
yanndupis: There are some discussions if we want to use `TensorShape` #499 across the code base. In the other layers we haven't use it until now.

PR #531, opened 2019-06-03
Support multiple outputs in the converter

yanndupis:
Hello, 

Currently, the converter doesn't support ops with multiple outputs when the op is not registered as a special op. Here is an attempt to fix this. By doing so, it should fix the issue for the `tf.split` operation identified in this pr #421.

Thanks,

### DIFF
--- a/tf_encrypted/convert/convert.py
+++ b/tf_encrypted/convert/convert.py
@@ -83,7 +83,7 @@ class Converter():
       if node.name not in specop_outputs:
 
         output = strip_tensor_info(node.name)
-        inputs = [strip_tensor_info(x) for x in node.input]
+        inputs = [x for x in node.input]
         if node.op == "Placeholder":
           try:
             _, item = inputs_iterable.__next__()
@@ -95,6 +95,18 @@ class Converter():
           continue
 
         self.outputs[output] = register[node.op](self, node, inputs)
+
+        out = register[node.op](self, node, inputs)
+
+        # if the operation returns a list with several ouputs,
+        # identify the outputs node name
+        if isinstance(out, list):
+          output_name = find_output_names(pb_trimmed, node.name)
+          for i, _ in enumerate(out):
+            self.outputs[output_name[i]] = out[i]
+        else:
+          self.outputs[output] = out
+
       else:
         # Register high level special operations
         for s in specop_dict:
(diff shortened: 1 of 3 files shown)


tf_encrypted/convert/convert.py
```diff
@@ -95,6 +95,18 @@ def convert(
           continue
 
         self.outputs[output] = register[node.op](self, node, inputs)
```
jvmancuso: Shouldn't this line be deleted now?

tf_encrypted/convert/convert.py
```diff
@@ -83,7 +83,7 @@ def convert(
       if node.name not in specop_outputs:
 
         output = strip_tensor_info(node.name)
-        inputs = [strip_tensor_info(x) for x in node.input]
+        inputs = [x for x in node.input]
```
jvmancuso: why this change?

tf_encrypted/convert/convert.py
yanndupis: Ah yes, my bad, it should have been removed.

tf_encrypted/convert/convert.py
yanndupis: ah it's because `strip_tensor_info` removes everything after `:` in the node name. However, at least for `split` operation, the outputs are distinguished by what's after `:`. For example, for split (in 3), it will be `split`, `split:1`, `split:2`. If we remove everything after `:`, the first output will be re-used three times.

jvmancuso: awesome, thanks for clarifying!

tf_encrypted/convert/convert.py
```diff
@@ -94,7 +94,17 @@ def convert(
           self.outputs[output] = x
           continue
 
-        self.outputs[output] = register[node.op](self, node, inputs)
+        out = register[node.op](self, node, inputs)
+
+        # if the operation returns a list with several ouputs,
+        # identify the outputs node name
+        if isinstance(out, list):
```
jvmancuso: can we switch this to `isinstance(out, (list, tuple))`? feels a bit more robust

tf_encrypted/convert/convert.py
```diff
@@ -293,3 +303,33 @@ def strip_tensor_info(node_name: str) -> str:
...
+
+  node {
+    name: "output/input"
+    op: "Pack"
+    input: "split"
+    input: "split:1"
+    }
+  """
+  output_node = []
+  for n in pb_trimmed:
```
jvmancuso: wish we could do this without iterating through the entire protobuf. I feel like we should be doing a forward lookahead from the current node that breaks when it stops encountering tensors that are outputs of the current node)

jvmancuso: this would assume that the nodes are ordered topologically, but if so that would be quite a bit faster

tf_encrypted/convert/convert.py
yanndupis: definitely, good point.

tf_encrypted/convert/convert.py
yanndupis: agreed, good point.

yanndupis: I tweaked the function so it doesn't go through the entire pb file. Let me know what you think. Thanks.

Issue #513, opened 2019-05-29
Module for secure randomness not loaded when installed from PyPI package


mortendahl:
TFE installed via pip will currently *not* load the secure randomness module. Error seems to come from TensorFlow:

```
>>> secure_random_module = tf.load_op_library(shared_object)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/miniconda3/envs/tfe-release-py35/lib/python3.5/site-packages/tensorflow/python/framework/load_library.py", line 61, in load_op_library
    lib_handle = py_tf.TF_LoadLibrary(library_filename)
tensorflow.python.framework.errors_impl.NotFoundError: dlopen(/usr/local/miniconda3/envs/tfe-release-py35/lib/python3.5/site-packages/tf_encrypted/operations/secure_random/secure_random_module_tf_1.13.1.so, 6): no suitable image found.  Did find:
	/usr/local/miniconda3/envs/tfe-release-py35/lib/python3.5/site-packages/tf_encrypted/operations/secure_random/secure_random_module_tf_1.13.1.so: unknown file type, first eight bytes: 0x7F 0x45 0x4C 0x46 0x02 0x01 0x01 0x00
	/usr/local/miniconda3/envs/tfe-release-py35/lib/python3.5/site-packages/tf_encrypted/operations/secure_random/secure_random_module_tf_1.13.1.so: unknown file type, first eight bytes: 0x7F 0x45 0x4C 0x46 0x02 0x01 0x01 0x00
```

Package should come with this for both Python 3.5 and 3.6 and both Linux and macOS.

justin1121:
Unable to reproduce on macos with python 3.5 and 3.6, linux with python 3.5 and 3.6. Blocked until we have a way to reproduce this reliably.

mortendahl:
Same here, I can't reproduce anymore. Closing down for now. Thanks a lot for looking into it @justin1121!

jvmancuso:
There are reports of PySyft users in the OpenMined Slack experiencing this error: https://openmined.slack.com/archives/C6EEFN3A8/p1559517593068600?thread_ts=1559517593.068600&cid=C6EEFN3A8

justin1121:
Others have had similar issues before https://github.com/bbalasub1/glmnet_python/issues/17. Even the first eight bytes is the same.

justin1121:
[...]

PR #545, opened 2019-06-05
Fix for TF v1.14.0-rc0

### LINKED ISSUE #526: 'base_layer_utils' has no attribute 'unique_layer_name'
AttributeError: module 'tensorflow.python.keras.engine.base_layer_utils' has no attribute 'unique_layer_name'.

Happened on TF 1.14.0rc0.

jvmancuso:
Fixes #526 

Previously, we were using `unique_layer_name` from tf.keras.engine.base_layer_utils, however in TensorFlow >1.13.0 this function was renamed to `unique_object_name` and moved to tf.keras.backend.py.  I've copied the function logic over, since it luckily only relies on common Python libraries and public TensorFlow functions (specifically, `tf.get_default_graph()`).


yanndupis:
@jvmancuso, looks great. If @mortendahl is ok I think., we can approve it. 

I have re-run the pysyft notebooks with this fix installed and TF 1.14.0-rc and it works. 

However I have noticed that I am getting the insecure randomness warning with this version of TF #513

```
WARNING: Logging before flag parsing goes to stderr.
W0605 12:49:04.146198 4654921152 secure_random.py:22] Falling back to insecure randomness since the required custom op could not be found for the installed version of TensorFlow (1.14.0-rc0). Fix this by compiling custom ops.
```

justin1121:
Yep there is work we need to do to be able to support other versions of tensorflow for secure numbers.

jvmancuso:
Cool, tagging #513 so there's a link back to this note in the relevant open issue

mortendahl:
it looks good, but one question: what are they doing in tf.keras if this function is gone? are we, or should we, be doing the same instead of using this older behaviour?

jvmancuso:
The behavior is identical: the only thing that's changed is the function name and it's location insider the TF codebase

Issue #550, opened 2019-06-06
Testing strategy for operations and their kernels


mortendahl:
Right now our tests for eg matmul is tied to a specific protocol, but would be mostly the same across various protocols. To avoid code duplication we need a way to have only one (base) test for matmul yet still run it against any protocol that claims to support it.

mortendahl:
Comment by @jvmancuso in https://github.com/tf-encrypted/tf-encrypted/pull/551:

> I think this PR reminds me of a few things we want to do (eventually phase out layers, break pond out into more files, rethink how a few generic functions are shared between tests) but I think that's all out of scope for this one!

PR #551, opened 2019-06-06
Move tests file next to implementation

### LINKED ISSUE #410: Move test files next to implementation files
Right now we have (almost) all tests files in the `/tests/` subdirectory, yet it seems more logical and convenient to keep them next to the corresponding code they're testing. This issue is about simply moving those files (with `git mv`) and patching any imports that may need it.

mortendahl:
Closes #410.

Some files were merged to have a single test file per implementation file. There were some questions around where to place a few of them but opened #550 to address this doing refactoring.

Note that I left many commits to better see what was done; these should be squeezed before merging.


mortendahl:
Added your comment to https://github.com/tf-encrypted/tf-encrypted/issues/550 @jvmancuso

PR #571, opened 2019-06-14
Convert tf.keras into tfe.keras model 

### LINKED ISSUE #552: Converting a tf.keras model into a tfe.keras model
### Context

TFE now provides a tfe.keras api which matches tf.keras. Once a public tf.keras model is trained, we would like to be able to easily convert it into tfe.keras model.

yanndupis:
Hey guys, 

The conversation is working but I would need your feedback to make it better.

Here is the existing API:
```
with tfe.protocol.SecureNN():
    
    tfe_model = tfe.keras.Sequential.from_config(config)
    
    x = tfe.define_private_variable(data)


with tfe.Session() as sess:
    sess.run(tf.global_variables_initializer())
    
    tfe_model.set_weights(keras_weights, sess)
    
    y = tfe_model(x)
    actual = sess.run(y.reveal())
```

closes #552


tf_encrypted/keras/models/sequential.py
```diff
@@ -101,3 +103,66 @@ def layers(self):
...
+
+    # Define keras weights as private variables
+    keras_weights_p = [tf_encrypted.define_private_variable(w)
+                       for w in keras_weights]
+
+    # Assign new keras weights to existing weights defined by
+    # default when tfe layer instantiated
+    sess.run(tf.global_variables_initializer())
+    for i, w in enumerate(model_weights):
+      sess.run(tf_encrypted.assign(w, keras_weights_p[i]))
```
yanndupis: @mortendahl  - couple question: 

- I am currently defining the keras weights  as  private variable in `set_weights` before assigning it to the tfe weights. I'd like to move this define_private_variable into `tfe.assign`, however I think I will need to have the Session in `tfe.assign` to initialize the weights `sess.run(tf.global_variables_initializer())`. Do you have any recommendation to approach this? 
- currently `set_weights` expects a tfe session. It's a bit different from the keras api. However I am not sure if  `get_tfe_session()` would work in this context.

tf_encrypted/keras/models/sequential.py
```diff
@@ -101,3 +103,66 @@ def layers(self):
...
+
+  @staticmethod
+  def from_config(keras_config):
```
yanndupis: currently to transform the keras model into tfe.keras model we have to do:
`tfe_model = tfe.keras.Sequential.from_config(config)`

Do we want also have the option to do:
- `tfe_model = tfe.keras.models.from_config(config)`

If yes, is it possible to add a static method to a module?

tf_encrypted/keras/layers/convolutional.py
```diff
@@ -157,12 +157,14 @@ def build(self, input_shape):
     self.kernel_shape = self.kernel_size + (input_dim, self.filters)
 
     kernel = self.kernel_initializer(self.kernel_shape)
-    self.kernel = self.prot.define_private_variable(kernel)
+    self.kernel = self.add_weight(
```
mortendahl: I'd expect any call to `define_private_variable` to happen in `add_weight` instead

tf_encrypted/keras/engine/base_layer.py
```diff
@@ -97,6 +99,12 @@ def __call__(self, inputs, *args, **kargs):
 
     return outputs
 
+  def add_weight(self, variable):
```
mortendahl: does it make sense if the input here is always a kernel?

tf_encrypted/keras/models/sequential.py
```diff
@@ -101,3 +103,66 @@ def layers(self):
...
+          matching the output of model.get_weights()
+      sess: tfe session"""
+
+    # List model weights
+    model_weights = []
+    for l in self.layers:
+      model_weights += l.weights
+
+    # Define keras weights as private variables
+    keras_weights_p = [tf_encrypted.define_private_variable(w)
```
mortendahl: we should not create new variables here, only assign to existing ones

tf_encrypted/keras/models/sequential.py
mortendahl: I think we should yes! as well as [`clone_model`](https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/keras/models/clone_model)

mortendahl: to add (static) methods to a module you simply define the function in a file:

```python
# in file tf_encrypted/keras/models.py
def clone_model(model, input_tensors=None):
  ...
```

tf_encrypted/keras/models/sequential.py
mortendahl: hmm, why would you create variables inside `tfe.assign`? it sounds like we're working around another issue here that should be fixed instead

examples/notebooks/keras_private_predict.ipynb
```diff
@@ -0,0 +1,178 @@
+{
```
mortendahl: <p>the call to `get_weights` and `get_config` seem more like an implementation detail, think the right thing to do is *not* to have them but instead call `tfe_model = tfe.keras.models.clone_from(model_k)` (see <a href="https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/keras/models/clone_model" target="_blank">https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/keras/models/clone_model</a>)</p>

 _Reply via <a href='https://app.reviewnb.com/tf-encrypted/tf-encrypted/pull/571/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

mortendahl: <p>using `clone_from` should allow us to avoid the call to `set_weights`</p>

 _Reply via <a href='https://app.reviewnb.com/tf-encrypted/tf-encrypted/pull/571/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

tf_encrypted/keras/engine/base_layer.py
jvmancuso: I'd say no, for example there are weights in batchnorm that are not kernels

examples/notebooks/keras_private_predict.ipynb
jvmancuso: <p>I agree that's cleaner, but I think we should support both. a user might want to do things like only set some of the weights (common in transfer learning)</p>

tf_encrypted/keras/models/sequential.py
jvmancuso: this should be implemented in the base Layer abstraction instead of at the model level (note that this class will inherit the method in that case, see the tf.keras source code for more)

tf_encrypted/keras/models/sequential.py
jvmancuso: good point

tf_encrypted/keras/models/sequential.py
yanndupis: How would you assign a numpy array to an existing private variable?

examples/notebooks/keras_private_predict.ipynb
yanndupis: I like the idea of adding `clone_from`. But let's keep the option to do `set_weights` and `get_config`.

tf_encrypted/keras/models/sequential.py
yanndupis: @mortendahl, when you say "yes" you mean `from config` should be a classmethod?

tf_encrypted/keras/models/sequential.py
yanndupis: should we use a placeholder? something like `sess.run(v.assign_from_same(w), feed_dict=p.feed(np.array([5, 5, 5])))`

https://github.com/tf-encrypted/tf-encrypted/blob/c9200f3d814f54128c4bad24c615458dfc8f253d/examples/deprecated/int100.py

tf_encrypted/keras/models/sequential.py
mortendahl: TF seems to have two calls for this, suspect one calls the other underneat:
- https://www.tensorflow.org/api_docs/python/tf/keras/models/model_from_config
- https://www.tensorflow.org/api_docs/python/tf/keras/models/Model#from_config

tf_encrypted/keras/models/sequential.py
yanndupis: in fact if use a placeholder we could something like:

```
new_weights_pl = tfe.define_private_placeholder(new_weights_shape)
fd = p.feed(new_weigths_array)
sess.run(tfe.assign(existing_variable_weights, new_weights_pl), feed_dict=fd)
```

mortendahl: yeah it'd probably have to be a placeholder, so that sharing happens locally. we did some code for this around the queues, maybe that fits directly!

examples/notebooks/keras_private_predict.ipynb
mortendahl: sounds good! this is a good target I think: https://www.tensorflow.org/api_docs/python/tf/keras/models

tf_encrypted/keras/models/sequential.py
mortendahl: ideally we'd move that placeholder into the tfe.Variable, so it can be re-used. problem with creating these objects (vars and placeholders) in the Keras code is that they add to the graph

tf_encrypted/keras/models/sequential.py
yanndupis: ah great, thanks for sharing this.

tf_encrypted/keras/models/sequential.py
yanndupis: Makes sense @mortendahl .

yanndupis: @jvmancuso good point, regarding `set_weights` in `base_layer`. Just added.

tf_encrypted/keras/models/sequential.py
yanndupis: `with tfe.Session() as sess` directly in `set_weights` to assign new weights doesn't give the right output. I guess because it's not re-using the same session when running the model with the following code:

```
with tfe.Session() as sess:
    sess.run(tf.global_variables_initializer())
    
    tfe_model.set_weights(keras_weights)
    
    y = tfe_model(x)
    actual = sess.run(y.reveal())
```

tf_encrypted/keras/engine/base_layer.py
```diff
@@ -97,6 +101,31 @@ def __call__(self, inputs, *args, **kargs):
...
+          matching the output of layer.get_weights()
+      sess: tfe session"""
+
+    # Define keras weights as private variable
+    tfe_weights_pl = [tfe.define_private_placeholder(w.shape)
+                      for w in weights]
+
+    # Assign new keras weights to existing weights defined by
+    # default when tfe layer was instantiated
+    sess = tf.get_default_session()
```
jvmancuso: this should use `K.get_session()` to be consistent

tf_encrypted/keras/models/sequential.py
```diff
@@ -101,3 +104,113 @@ def layers(self):
...
+      weights: A list of Numpy arrays with shapes and types
+          matching the output of model.get_weights()
+      sess: tfe session"""
+
+    # Updated weights for each layer
+    for layer in self.layers:
+      num_param = len(layer.weights)
+      layer_weights = weights[:num_param]
+      # Define keras weights as private placeholder
+      tfe_weights_pl = [tfe.define_private_placeholder(w.shape)
```
jvmancuso: I'm pretty sure everything from here through line 128 can be replaced by `layer.set_weights(layer_weights, sess)`

tf_encrypted/keras/models/sequential.py
```diff
@@ -101,3 +104,113 @@ def layers(self):
...
+      # default when tfe layer was instantiated
+      if not sess:
+        sess = K.get_session()
+      for i, w in enumerate(layer.weights):
+        fd = tfe_weights_pl[i].feed(layer_weights[i])
+        sess.run(tfe.assign(w, tfe_weights_pl[i]), feed_dict=fd)
+
+      weights = weights[num_param:]
+
+  def from_config(self, config):
```
jvmancuso: this should be a classmethod

tf_encrypted/keras/models/sequential.py
```diff
@@ -101,3 +104,113 @@ def layers(self):
...
+
+  Arguments:
+      config: Configuration dictionary.
+          matching the output of model.get_weights()
+      sess: tfe session
+
+  Returns:
+      A TFE Keras model instance
+  """
+  return _rebuild_tfe_model(config)
```
jvmancuso: I would write the logic for `_rebuild_tfe_model` here explicitly, there's no real reason to break it out

tf_encrypted/keras/models/sequential.py
```diff
@@ -101,3 +104,113 @@ def layers(self):
...
+
+    Arguments:
+        config: Configuration dictionary.
+            matching the output of model.get_weights()
+        sess: tfe session
+
+    Returns:
+        A TFE Keras model instance
+    """
+    tfe_model = _rebuild_tfe_model(config)
```
jvmancuso: I'd suggest calling `model_from_config` here

tf_encrypted/keras/models/sequential.py
```diff
@@ -101,3 +104,113 @@ def layers(self):
...
+  tfe_model = tfe.keras.Sequential([])
+
+  for k_l_c in keras_config['layers']:
+    tfe_layer = _instantiate_tfe_layer(k_l_c)
+    tfe_model.add(tfe_layer)
+
+  return tfe_model
+
+def _instantiate_tfe_layer(keras_layer_config):
+  """instantiate tfe layer based on layer keras config"""
```
jvmancuso: please make this a Google style docstring :)

examples/notebooks/b - Secure Classification with TFE Keras - Secure Model Serving .ipynb
```diff
@@ -0,0 +1,302 @@
+{
```
jvmancuso: <p>the first `global request_ix` can be removed</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tf-encrypted/tf-encrypted/pull/571/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

examples/notebooks/c - Secure Classification with TFE Keras - Private Prediction Client.ipynb
```diff
@@ -0,0 +1,154 @@
+{
```
jvmancuso: <p>nitpick: wrongly --&gt; incorrectly :)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tf-encrypted/tf-encrypted/pull/571/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

examples/notebooks/keras_private_predict.ipynb
```diff
@@ -0,0 +1,365 @@
...
```
jvmancuso: <p>is this notebook still useful here?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tf-encrypted/tf-encrypted/pull/571/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

yanndupis: Good point, it should be removed.

tf_encrypted/keras/models/sequential.py
yanndupis: ah correct, I am sure we can also do this :)

tf_encrypted/keras/models/sequential.py
yanndupis: Good point, let's do that. thanks

examples/notebooks/c - Secure Classification with TFE Keras - Private Prediction Client.ipynb
yanndupis: ah thanks

yanndupis:
I think it looks much better now, thanks for your review @jvmancuso!

examples/notebooks/b - Secure Classification with TFE Keras - Secure Model Serving .ipynb
```diff
@@ -0,0 +1,275 @@
...
```
jvmancuso: <p>Sorry one last thing! Shouldn't `num_steps=3` here?  Otherwise the last cell will hang until it serves two more prediction requests.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tf-encrypted/tf-encrypted/pull/571/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

jvmancuso: either that or num_tests in the other notebook should be 5

yanndupis: ah good catch, let me fix this.

PR #578, opened 2019-06-19
MNIST example using Keras API

VMS-6511:
This PR is currently blocked by the open PR #571 so that the weights can be loaded in.


examples/mnist/run.py
```diff
@@ -137,6 +137,7 @@ def normalize(image, label):
 
     iterator = dataset.make_one_shot_iterator()
     return iterator
+
```
jvmancuso: delete this

examples/mnist/run.py
```diff
@@ -177,15 +177,24 @@ def receive_output(self, logits: tf.Tensor) -> tf.Operation:
...
 
+  w0, b0, w1, b1 = params
+  initializer_w0 = tf.keras.initializers.Constant(w0)
+  initializer_b0 = tf.keras.initializers.Constant(b0)
+  initializer_w1 = tf.keras.initializers.Constant(w1)
+  initializer_b1 = tf.keras.initializers.Constant(b1)
+  model = tfe.keras.Sequential([
+    tfe.keras.layers.Dense(512, batch_input_shape=(PredictionClient.BATCH_SIZE, 784), kernel_initializer=initializer_w0, 
+    bias_initializer=initializer_b0),
+    tfe.keras.layers.BatchNormalization(axis=1),
```
jvmancuso: how does this perform normalization if the batch statistics weights haven't been initialized?  we can't add layers here without changing the way training happens in `ModelOwner._build_training_graph`

jvmancuso:
also, have you run these examples to make sure they still return the right answers?

examples/mnist/run.py
VMS-6511: Agree with this, had removed it from my other implementations but forgot to here. Thanks, will do so!

VMS-6511:
I haven't yet run the examples since I can't initialize the weights but I will once #571 is merged.

jvmancuso:
you should be able to with the way you're doing it now (using kernel initializers manually), it's just not how we'd ideally like to present it in the examples.  but it's probably okay to hold off since #571 should be merged soon

examples/mnist/run.py
```diff
@@ -57,43 +64,33 @@ def normalize(image, label):
...
-    # training loop
-    def loop_body(i):
-
-      # get next batch
-      x, y = training_data.get_next()
+    model = keras.Sequential()
+    model.add(keras.layers.Flatten())
+    model.add(keras.layers.Dense(512, input_shape=[28, 28, 1]))
+    model.add(keras.layers.Activation('relu'))
+    model.add(keras.layers.Dense(self.NUM_CLASSES, activation='softmax'))
```
jvmancuso: the softmax here should be removed (the loss function you're using expects logits and not probabilities)

examples/mnist/run.py
```diff
@@ -170,22 +166,27 @@ def receive_output(self, logits: tf.Tensor) -> tf.Operation:
       local_data_file="./data/test.tfrecord")
 
   # get model parameters as private tensors from model owner
-  params = tfe.define_private_input(model_owner.player_name,
-                                    model_owner.provide_input, masked=True)  # pylint: disable=E0632
+  params = tfe.define_private_variable(model_owner.player_name,
```
jvmancuso: I think it's okay to use `define_private_input` here

jvmancuso: sorry, I know I kept mentioning private_variable but the way @yanndupis 's code worked out it makes sense both ways and this one is more convenient because it can do masking automatically 😄

examples/mnist/run.py
```diff
@@ -170,22 +166,27 @@ def receive_output(self, logits: tf.Tensor) -> tf.Operation:
...
 
   # we'll use the same parameters for each prediction so we cache them to
   # avoid re-training each time
   cache_updater, params = tfe.cache(params)
 
+  w0, b0, w1, b1 = params
+  model = tfe.keras.Sequential([
+    tfe.keras.layers.Dense(512, batch_input_shape=(PredictionClient.BATCH_SIZE, 784)),
+    tfe.keras.layers.Activation('relu'),
+    tfe.keras.layers.Dense(10, activation=softmax)
```
jvmancuso: same comment as above, no softmax

examples/mnist/run.py
```diff
@@ -170,22 +166,27 @@ def receive_output(self, logits: tf.Tensor) -> tf.Operation:
...
-  params = tfe.define_private_input(model_owner.player_name,
-                                    model_owner.provide_input, masked=True)  # pylint: disable=E0632
+  params = tfe.define_private_variable(model_owner.player_name,
+                                    model_owner.provide_input)  # pylint: disable=E0632
 
   # we'll use the same parameters for each prediction so we cache them to
   # avoid re-training each time
   cache_updater, params = tfe.cache(params)
 
+  w0, b0, w1, b1 = params
```
jvmancuso: you can remove this line and just provide do `model.set_weights(params)` below

examples/mnist/run.py
```diff
@@ -7,10 +7,15 @@
 import sys
 
 import tensorflow as tf
+from tensorflow import contrib
```
jvmancuso: this line is unused?

examples/mnist/run.py
```diff
@@ -57,43 +64,33 @@ def normalize(image, label):
...
+    def loss(model, x, y):
+      y_ = model(x)
+      return tf.losses.sparse_softmax_cross_entropy(labels=y, logits=y_)
 
-      predictions = layer2
-      loss = tf.reduce_mean(
-          tf.losses.sparse_softmax_cross_entropy(logits=predictions, labels=y))
-      with tf.control_dependencies([optimizer.minimize(loss)]):
-        return i + 1
+    def grad(model, inputs, targets):
```
jvmancuso: this function can be switched to:
```python
def grad(model, inputs, targets):
    loss_value = loss(model, inputs, targets)
    return loss_value, tf.gradients(loss_value, model.trainable_variables)
```

examples/mnist/run.py
```diff
@@ -57,43 +64,33 @@ def normalize(image, label):
...
+    model.add(keras.layers.Activation('relu'))
+    model.add(keras.layers.Dense(self.NUM_CLASSES, activation='softmax'))
 
-      # model construction
-      layer0 = tf.matmul(x, w0) + b0
-      layer1 = tf.nn.sigmoid(layer0)
-      layer2 = tf.matmul(layer1, w1) + b1
+    def loss(model, x, y):
+      y_ = model(x)
+      return tf.losses.sparse_softmax_cross_entropy(labels=y, logits=y_)
```
jvmancuso: this should also go through a `tf.reduce_mean` before being used in the `grad` function below

examples/mnist/run.py
```diff
@@ -57,43 +64,33 @@ def normalize(image, label):
...
+      return loss_value, tape.gradient(loss_value, model.trainable_variables)
 
-    loop = tf.while_loop(lambda i: i < self.ITERATIONS
-                         * self.EPOCHS, loop_body, (0,))
+    # optimizer and data pipeline
+    optimizer = tf.train.AdamOptimizer(learning_rate=0.01)
+    global_step = tf.Variable(0)
+    for epoch in range(self.EPOCHS):
+      # Training loop - using batches of 32
+      for x, y in training_data.get_next():
```
jvmancuso: I'm not sure that training_data.get_next() is a generator so I don't think this will work as intended -- I'd revert to something more similar to what we had before:
```python
def loop_body(i):
    x, y = training_data.get_next()
    loss_value, grads = grad(model, x, y)
    update_op = optimizer.apply_gradients(zip(grads, model.trainable_variables), global_step)
    with tf.control_dependencies([update_op]):
        return global_step
```
then you can use the same `tf.while_loop` code we had before, as well as everything after it but returning the `model.trainable_weights` instead of the list

VMS-6511:
Sorry, missed the above comments, will make the changes and push again.

examples/mnist/run.py
```diff
@@ -170,22 +168,22 @@ def receive_output(self, logits: tf.Tensor) -> tf.Operation:
       local_data_file="./data/test.tfrecord")
 
   # get model parameters as private tensors from model owner
-  params = tfe.define_private_input(model_owner.player_name,
-                                    model_owner.provide_input, masked=True)  # pylint: disable=E0632
-
-  # we'll use the same parameters for each prediction so we cache them to
-  # avoid re-training each time
+  params = [tfe.define_private_variable(x) for x in model_owner.provide_input()]  # pylint: disable=E0632
```
jvmancuso: this change should be reverted

examples/mnist/run.py
```diff
@@ -57,51 +64,43 @@ def normalize(image, label):
...
-    # return model parameters after training
-    with tf.control_dependencies([loop]):
-      print_op = tf.print("Training complete")
-    with tf.control_dependencies([print_op]):
-      return [w0.read_value(),
-              b0.read_value(),
-              w1.read_value(),
-              b1.read_value()]
+    
+    return model.get_weights()
```
jvmancuso: this should be returning `model.trainable_variables` or `model.trainable_weights`, since `model.get_weights()` returns numpy arrays, which we don't want

examples/mnist/run.py
VMS-6511: I switched it to `define_private_variable` in the most recent commit as I encountered this error with `define_private_input`:

```W
  File "run.py", line 169, in <module>
    params = tfe.define_private_input(model_owner.player_name,model_owner.provide_input)
  File "/Users/vinithms/Desktop/Dropout Labs/tf-encrypted/tf_encrypted/protocol/pond/pond.py", line 570, in define_private_input
    factory=factory,
  File "/Users/vinithms/Desktop/Dropout Labs/tf-encrypted/tf_encrypted/protocol/pond/pond.py", line 534, in define_local_computation
    return [share_output(output) for output in outputs]
  File "/Users/vinithms/Desktop/Dropout Labs/tf-encrypted/tf_encrypted/protocol/pond/pond.py", line 534, in <listcomp>
    return [share_output(output) for output in outputs]
  File "/Users/vinithms/Desktop/Dropout Labs/tf-encrypted/tf_encrypted/protocol/pond/pond.py", line 476, in share_output
    enc = self._encode(v, apply_scaling, tf_int_type=factory.native_type)
  File "/Users/vinithms/Desktop/Dropout Labs/tf-encrypted/tf_encrypted/protocol/pond/pond.py", line 640, in _encode
    assert type(rationals) == type(integers)  # pylint: disable=unidiomatic-typecheck
AssertionError```

Comment (jvmancuso) [reply examples/mnist/run.py]: I think that bug is a result of this problem: https://github.com/tf-encrypted/tf-encrypted/pull/578#discussion_r297451311

File: examples/mnist/run.py (hunk 12 above)
Comment (VMS-6511) [reply examples/mnist/run.py]: I've been encountering this error: ```Traceback (most recent call last):
  File "run.py", line 169, in <module>
    params = [tfe.define_private_variable(x) for x in model_owner.provide_input()]  # pylint: disable=E0632
  File "run.py", line 169, in <listcomp>
    params = [tfe.define_private_variable(x) for x in model_owner.provide_input()]  # pylint: disable=E0632
  File "/Users/vinithms/Desktop/Dropout Labs/tf-encrypted/tf_encrypted/protocol/pond/pond.py", line 334, in define_private_variable
    assert isinstance(initial_value, init_val_types), type(initial_value)
AssertionError: <class 'tensorflow.python.ops.resource_variable_ops.ResourceVariable'>```

File: examples/mnist/run.py [hunk 13]
```diff
@@ -57,51 +64,43 @@ def normalize(image, label):
...
 
-    # model parameters and initial values
-    w0 = tf.Variable(tf.random_normal([28 * 28, 512]))
-    b0 = tf.Variable(tf.zeros([512]))
-    w1 = tf.Variable(tf.random_normal([512, 10]))
-    b1 = tf.Variable(tf.zeros([10]))
-
+    model = keras.Sequential()
+    model.add(keras.layers.Flatten())
+    model.add(keras.layers.Dense(512, input_shape=[28, 28, 1]))
```
yanndupis: Why do we have Flatten layer before a Dense layer with input_shape [28, 28, 1]?

yanndupis: oups realized I had this in the notebook example... because `input_shape` is in the second layer, it gets ignored. that's why it can run even if the input shape after Flatten() is not [?, 28, 28, 1] but [?, 784]. I would recommend removing `input_shape=[28, 28, 1]` from the Dense layer or move it to the Flatten() layer.

jvmancuso: Or better yet, move the reshaping logic to the data pipeline function so that the data is already flattened by the time it gets here

yanndupis: agreed, that would be better.

examples/mnist/run.py
```diff
@@ -170,22 +168,22 @@ def receive_output(self, logits: tf.Tensor) -> tf.Operation:
...
-  params = tfe.define_private_input(model_owner.player_name,
-                                    model_owner.provide_input, masked=True)  # pylint: disable=E0632
-
-  # we'll use the same parameters for each prediction so we cache them to
-  # avoid re-training each time
+  params = [tfe.define_private_variable(x) for x in model_owner.provide_input()]  # pylint: disable=E0632
+  
   cache_updater, params = tfe.cache(params)
 
+  model = tfe.keras.Sequential()
```
yanndupis: we should define the model under `with tfe.protocol.SecureNN():` in order to use the exact relu instead of the approximated one.

examples/mnist/run.py
```diff
@@ -168,24 +164,25 @@ def receive_output(self, logits: tf.Tensor) -> tf.Operation:
...
-  w0, b0, w1, b1 = params
-  layer0 = tfe.matmul(x, w0) + b0
-  layer1 = tfe.sigmoid(layer0 * 0.1)  # input normalized to avoid large values
-  logits = tfe.matmul(layer1, w1) + b1
+  with tfe.protocol.SecureNN():
+    model = tfe.keras.Sequential()
+    model.add(tfe.keras.layers.Dense(512, input_shape=[PredictionClient.BATCH_SIZE,28*28]))
+    model.add(tfe.keras.layers.Activation('relu'))
+    model.add(tfe.keras.layers.Dense(10, activation=None))
+    model.set_weights(params)
```
yanndupis: hmm, we might have to move `set_weights` under `with tfe.Session(target=session_target)` to make sure it's using the tfe session when assigning variables:
https://github.com/tf-encrypted/tf-encrypted/blob/9bda5c593aab39a599b5e2634a6e21e7bf01ee83/tf_encrypted/keras/engine/base_layer.py#L128-L138

examples/mnist/run.py
```diff
@@ -167,13 +167,13 @@ def receive_output(self, logits: tf.Tensor) -> tf.Operation:
   
   # get model parameters as private tensors from model owner
   params = tfe.define_private_input(model_owner.player_name, model_owner.provide_input) # pylint: disable=E0632
+
```
VMS-6511: Will remove this newline.

PR #579, opened 2019-06-19
Fix converter when the last op returns a list of Tensors

### LINKED ISSUE #561: tf.split in keras custom layer
Hey @yanndupis 
this is a follow up to the bug I raised in #421. I use tf.split in a keras custom layer - 
```
class customLayer(tf.keras.layers.Layer):
    def __init__(self):
        super(customLayer, self).__init__()

    def build(self, input_shape):
        pass

    def call(self, input, **kwargs):
        return tf.split(input, 3, axis=-1)
```

I use this layer in the following simple model and then I convert it using secure_model
```
input_img = tf.keras.layers.Input(shape=(10,10,3))
x = customLayer()(input_img)
model = tf.keras.Model(inputs=input_img, outputs=x)
model.compile('adam', 'categorical_crossentropy', metrics=['accuracy'])
s_model = secure_model(model)
```

This raises the following error - 

Traceback (most recent call last):
  File "/cs/labs/peleg/avitalsh/CFNAS/CFNAS/test.py", line 33, in <module>
    s_model = secure_model(model)
  File "/cs/labs/peleg/avitalsh/tools/tfe_fork/tf-encrypted/tf_encrypted/private_model.py", line 84, in secure_model
    tfe.convert.registry(), 'input-provider', inputs)
  File "/cs/labs/peleg/avitalsh/tools/tfe_fork/tf-encrypted/tf_encrypted/convert/convert.py", line 104, in convert
    self.outputs[output_name[i]] = out[i]
IndexError: list index out of range

I tried debugging a bit and this is what I found - 
When the converter reach this line - 
https://github.com/tf-encrypted/tf-encrypted/blob/10218935bd7433775f7e0cfc27666881a300f028/tf_encrypted/convert/convert.py#L102

we enter ```find_output_names``` where we list the node names - 
https://github.com/tf-encrypted/tf-encrypted/blob/10218935bd7433775f7e0cfc27666881a300f028/tf_encrypted/convert/convert.py#L327-L328

and this is what I get - 
image

then when we iterate the list starting from index n_i + 1 which is 3 in this example - but ```node_name_list``` has only 3 items so this for loop is skipped and we end up with an empty list as a return value so ```output_name``` is an empty list
https://github.com/tf-encrypted/tf-encrypted/blob/10218935bd7433775f7e0cfc27666881a300f028/tf_encrypted/convert/convert.py#L102

and then we get the error here, trying to access items in an empty list
https://github.com/tf-encrypted/tf-encrypted/blob/10218935bd7433775f7e0cfc27666881a300f028/tf_encrypted/convert/convert.py#L103-L104

The converter works fine with custom keras layers which use operations that only have one output (such as tf.keras.layers.Relu and tf.keras.layers.Conv2D)

yanndupis:
Closes #561 

The converter is crashing when the last node returns a list with of Tensors. It worked properly when the the op with multiple outputs is not the last op. The test for split was working because the list of tensors was getting packed with tf.identity in `export`. So in the test, tf.split wasn't the last node. 

Thanks,

### DIFF
--- a/tf_encrypted/convert/convert_test.py
+++ b/tf_encrypted/convert/convert_test.py
@@ -888,11 +888,7 @@ def export(x: tf.Tensor, filename: str, sess=None):
     sess = tf.Session()
 
   pred_node_names = ["output"]
-  if isinstance(x, (list, tuple)):
-    x = tf.identity(x[0], name=pred_node_names[0])
-  else:
-    tf.identity(x, name=pred_node_names[0])
-
+  tf.identity(x, name=pred_node_names[0])
   graph = graph_util.convert_variables_to_constants(
       sess,
       sess.graph.as_graph_def(),
--- a/tf_encrypted/convert/convert_test.py
+++ b/tf_encrypted/convert/convert_test.py
@@ -638,9 +638,7 @@ def run_split(data):
 
 def export_split(filename, input_shape):
   a = tf.placeholder(tf.float32, shape=input_shape, name="input")
-  # re-name the op as output, so the list of Tensors in the output
-  # doens't get Packed by tf.identity in the export function
-  x = tf.split(a, num_or_size_splits=3, axis=-1, name="output")
+  x = tf.split(a, num_or_size_splits=3, axis=-1)
   return export(x, filename)
 
 
@@ -889,16 +887,11 @@ def export(x: tf.Tensor, filename: str, sess=None):
     should_close = True
     sess = tf.Session()
 
-  last_node_name = sess.graph.as_graph_def().node[-1].name
-
-  # if the op returns a list of tensors, we want avoid
-  # adding a tf.identiy op so the the list of tensors doesn't
-  # get packed.
-  if last_node_name != "output":
-    pred_node_names = ["output"]
-    tf.identity(x, name=pred_node_names[0])
+  pred_node_names = ["output"]
+  if isinstance(x, (list, tuple)):
+    x = tf.identity(x[0], name=pred_node_names[0])
   else:
-    pred_node_names = [last_node_name]
+    tf.identity(x, name=pred_node_names[0])
 
   graph = graph_util.convert_variables_to_constants(
       sess,
--- a/tf_encrypted/convert/convert_test.py
+++ b/tf_encrypted/convert/convert_test.py
@@ -638,7 +638,9 @@ def run_split(data):
 
 def export_split(filename, input_shape):
   a = tf.placeholder(tf.float32, shape=input_shape, name="input")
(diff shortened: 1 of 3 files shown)


tf_encrypted/convert/convert_test.py
```diff
@@ -887,8 +889,17 @@ def export(x: tf.Tensor, filename: str, sess=None):
...
-  pred_node_names = ["output"]
-  tf.identity(x, name=pred_node_names[0])
+  last_node_name = sess.graph.as_graph_def().node[-1].name
+
+  # if the op returns a list of tensors, we want avoid
+  # adding a tf.identiy op so the the list of tensors doesn't
+  # get packed.
+  if last_node_name != "output":
+    pred_node_names = ["output"]
+    tf.identity(x, name=pred_node_names[0])
```
jvmancuso: hm, would it be easier if we just check to see if x is a list instead of a single tensor, and then do `out = [tf.identity(xi) for xi in x]`?

jvmancuso: perhaps I'm misunderstanding what the problem is here though

yanndupis: Just changed  the approach a bit. However I am getting an error for special ops: 'required_space_to_batch_paddings':

```
def _register_specop(self, node, specop_scope_dict):
      """Handle special op registration."""
      input_list = specop_scope_dict['inputs']
      output_list = specop_scope_dict['outputs']
    
      # Handle edge cases if the ops return multiple outputs
      op_handler = self.registry[specop_scope_dict['op']]
    
      nodes = specop_scope_dict['interiors']
      if not nodes:
        nodes = node
      outs = op_handler(self, nodes, input_list)
      if isinstance(outs, (list, tuple)):
        for i, x in enumerate(outs):
>         self.outputs[output_list[i]] = x
E         IndexError: list index out of range

tf_encrypted/convert/convert.py:147: IndexError
```

`tf.identity` was also probably packing the outputs from `required_space_to_batch_paddings` that's why probably we haven't seen this bug :)

yanndupis: Ok... so the problem is not the converter, it's the test. `required_space_to_batch_paddings` output two nodes in the graph: padding and crop. Whereas tf.split is always the last node. I tried in `graph_util.convert_variables_to_constants` to list two nodes as output but it doesn't work properly when the graph gets exported. 

I see two solution for now:
- we keep the original test with `tf.identity` it's just that the output for `required_space_to_batch_paddings`  and `split` will get packed. or
- we use the second version of the test where I renamed the node split as output. In this case the output for split won't be packed but, we will leave the output for `required_space_to_batch_paddings`. but the code looks less clean.

PR #583, opened 2019-06-24
Keras batchnorm convert

avitalsh:
Added keras batchnorm conversion and test. 
Included batchnorm as a special op.
2 small changes to convert:
1. in the **get_interiors** method -  after finding one match, we don't need to continue looking. This causes an error when there are sub-nodes of the searched node. For example, in ```batch_normalization_v1``` we are looking for the ```FusedBatchNorm``` op. The correct node would be: ```batch_normalization_v1/cond/FusedBatchNorm```. But this node has sub nodes such as ```batch_normalization_v1/cond/FusedBatchNorm/Switch``` and without using break, we will end up with the sub node instead of the correct node.

2. in the **find_leaves** method -  in order to find leaves, we remove each node's inputs from the ```output_leaves``` list. I've added support for the case that the input is of the form: _nodename_:1. For example, in ```batch_normalization_v1``` there is a node named: ```batch_normalization_v1/cond/Switch``` which is not removed from the ```output_leaves``` list because it is used as input to other nodes as follows: ```batch_normalization_v1/cond/Switch:1```.  

**Problem with conversion test:**
Currently the test is failing. I've tested the conversion using my own code (created a model, converted to secure_model and compared prediction results) and it works.
I also tried running the same logic as in the conversion test **outside**  convert_test and it works fine. 
The problem is with the graph def - there are two nodes that from some reason are not "named" under batchnormalization. They are named:
```keras_learning_phase/input``` and ```keras_learning_phase```
but they should be named:
```batch_normalization_v1/keras_learning_phase/input``` and ```batch_normalization_v1/keras_learning_phase```
because of this they are not removed from the node list in the convert method, as the rest of the sub-nodes.  
I'm not sure why this is happening and would appreciate any help to resolve this issue.

### CODE REVIEW

File: tf_encrypted/convert/register.py [hunk 1]
```diff
@@ -198,6 +199,33 @@ def _keras_dense(converter, interiors, inputs):
...
+  fmt = bn_op.attr["data_format"].s.decode('ascii')
+
+  gamma = _nodef_to_numpy_array(interiors["gamma"])
+  beta = _nodef_to_numpy_array(interiors["beta"])
+  moving_mean = _nodef_to_numpy_array(interiors["moving_mean"])
+  moving_variance = _nodef_to_numpy_array(interiors["moving_variance"])
+
+  input_shape = x_in.shape.as_list()
+
+  layer = Batchnorm(input_shape=input_shape,
```
jvmancuso: can we use the `tfe.keras.layers.BatchNormalization` layer here instead?  we're slowly working on phasing out the tfe.layers module

avitalsh:
@jvmancuso I changed the batchnorm layer to be tfe.keras.layers instead of tfe.layers. 
The failing test issue still exists, do you have any idea why this is happening? As I've mentioned, when I run the same logic outside the test, the graph_def is fine and the ```keras_learning_phase/input``` and ```keras_learning_phase``` nodes are removed from the node list.

avitalsh:
Hi @jvmancuso , did you get a chance to look into this?

jvmancuso:
I've been looking into it on and off and I'm not exactly sure why that scope is changing for the test only. I will give it another shot when I can, but let me know if you have any idea why the scopes are different

jvmancuso:
what's interesting about this problem is that others have reported the opposite behavior: https://stackoverflow.com/questions/46259467/tensorboard-and-dropout-layers/49757651#49757651

avitalsh:
What bothers me is why this is happening only in the test. I tried the following in a different place and the graph def is fine - 
```
from tf_encrypted.convert.convert_test import _keras_batchnorm_core
model, _ = _keras_batchnorm_core(shape=[1, 28, 28, 1])
sess = K.get_session()
```
But when looking at sess.graph_def at ```export_keras_batchnorm``` (when debugging the test) - the issue exists. (I also tried calling ```export_keras_batchnorm``` from a different script and it works fine)

jvmancuso:
Ah, I think I found it -- if you remove `K.clear_session()` from the test case's `setUp` method, those ops are created under the correct scope.  My interpretation of this is that calling `clear_session()` preemptively resets the learning phase, which adds nodes to the graph before any other scopes have been created in it, and then the BatchNormalization layer checks to see if that Placeholder already exists, in which case it skips creating it under its own scope.

After fixing this, I ran into another bug in `_nodef_to_numpy_array`, I think it was a silent bug that was somehow masked before.  When passing `interiors['gamma']` to `_nodef_to_numpy_array`, I get the following error:
```python
x = name: "batch_normalization_v1/gamma"
op: "Const"
attr {
  key: "dtype"
  value {
    type: DT_FLOAT
  }
}
attr {
  key...type: DT_FLOAT
      tensor_shape {
        dim {
          size: 1
        }
      }
      float_val: 1.0
    }
  }
}


    def _nodef_to_numpy_array(x):
      """Map a NodeDef x to a np.array."""
      dtype = x.attr["dtype"].type
      print("dtype", dtype)
      x_shape = [i.size for i in x.attr["value"].tensor.tensor_shape.dim]
    
      if dtype == tf.float32:
        print("check1")
        nums = array.array('f', x.attr["value"].tensor.tensor_content)
      elif dtype == tf.float64:
        print("check2")
        nums = array.array('d', x.attr["value"].tensor.tensor_content)
      elif dtype == tf.int32:
        print("check3")
        nums = array.array('i', x.attr["value"].tensor.tensor_content)
      else:
        print("check4")
        raise TypeError("Unsupported dtype")
    
>     return np.array(nums).reshape(x_shape)
E     ValueError: cannot reshape array of size 0 into shape (1,)
tf_encrypted/convert/register.py:785: ValueError
------------------------------------------------------------------------------------------------------------ Captured stdout call ------------------------------------------------------------------------------------------------------------
dtype 1
check1
```

avitalsh:
@jvmancuso Great!
looking into the new bug now. This does not happen when the model is defined by tf.keras.Model:
```
input_img = tf.keras.layers.Input(shape=(10,10,3))
x = tf.keras.layers.BatchNormalization()(input_img)
model = tf.keras.Model(inputs=input_img, outputs=x)
```
In this case, when When passing ```interiors['gamma']``` to ```_nodef_to_numpy_array``` the input node looks like this - 
```
name: "batch_normalization_v1/gamma"
op: "Const"
attr {
  key: "dtype"
  value {
    type: DT_FLOAT
  }
}
attr {
  key: "value"
  value {
    tensor {
      dtype: DT_FLOAT
      tensor_shape {
        dim {
          size: 3
        }
      }
      tensor_content: "\000\000\200?\000\000\200?\000\000\200?"
    }
  }
}
```

But when using Sequential model - 
```
model = tf.keras.Sequential()
bn = tf.keras.layers.BatchNormalization(input_shape=[28, 28, 1])
model.add(bn)
```

the input node looks like this - 
```
name: "batch_normalization_v1/gamma"
op: "Const"
attr {
  key: "dtype"
  value {
    type: DT_FLOAT
  }
}
attr {
  key: "value"
  value {
    tensor {
      dtype: DT_FLOAT
      tensor_shape {
        dim {
          size: 1
        }
      }
      float_val: 1.0
    }
  }
}
```

without ```tensor_content``` and wrong shape.

I'm trying to understand why, if you have any idea please let me know.

jvmancuso:
I haven't been able to verify it yet, but I'm suspicious that the problem might be related to the behavior of fused vs. not fused in BatchNormalizationV1 versus V2.

jvmancuso:
in this case, I think it's okay to just check if `x.attr["value"].tensor.tensor_content is None` inside that if branch, and if it is, try to use `x.attr["value"].tensor.float_val` instead

tf_encrypted/convert/register.py
```diff
@@ -50,6 +51,7 @@ def registry():
       'Identity': _identity,
       "GatherV2": _gather,
       "dense": _keras_dense,
+      "batch_normalization_v1": _keras_batchnorm,
```
jvmancuso: will this work if the user supplies the V2 version of keras's batchnorm?

jvmancuso: it's okay if it won't, we can leave that for another PR, just curious!

avitalsh: I'm not sure yet, trying to understand how to call V2 - in my tf version (1.13.1) batchnorm layer always runs V1, and V2 is for tensorflow 2.0
```
if tf2.enabled():
  BatchNormalization = BatchNormalizationV2
else:
  BatchNormalization = BatchNormalizationV1
```

But perhaps newer tf versions do call V2?

jvmancuso: ah, that would make sense.  it looked like at one point during 1.xx they added a V2 of batchnorm, but I wasn't sure if that was the same difference that I was seeing here.  thanks!

PR #584, opened 2019-06-24
make sure tfe.keras.layers use same tf.keras.layers default args

### LINKED ISSUE #582: dilation_rate not implemented error in keras conv2D
Hi, 
I was trying to use the new tf.keras to tfe.keras conversion (PR #571 ) using the following code - 
```
model = tf.keras.Sequential()
model.add(tf.keras.layers.Conv2D(filters=16, kernel_size=(3,3), use_bias=False))

model.compile('adam', 'categorical_crossentropy', metrics=['accuracy'])

config = tfe.LocalConfig()
tfe.set_config(config)
tfe.set_protocol(tfe.protocol.SecureNN())

tf.reset_default_graph()

with tfe.protocol.SecureNN():
    tfe_model = tfe.keras.models.clone_model(model)
```

```tf.keras.layers.Conv2D``` has a default ```dilation_rate``` value (1,1). While building the tfe.keras model we reach this if statement:
https://github.com/tf-encrypted/tf-encrypted/blob/0c4aa9b5a2a16b44e4d595e73b72c11d02a90140/tf_encrypted/keras/layers/convolutional.py#L122-L125
Which makes us crush with the following error - 
NotImplementedError: `dilation_rate` argument is not implemented for layer Conv2d
 
I tried setting ```dilation_rate```  to None as a quick workaround but it raised a value error because ```dilation_rate``` should be: 

>  an integer or tuple/list of 2 integers

### PR DESCRIPTION
**Pull request by yanndupis**
Closed #582

### DIFF
--- a/tf_encrypted/keras/engine/base_layer.py
+++ b/tf_encrypted/keras/engine/base_layer.py
@@ -115,17 +115,15 @@ class Layer(ABC):
           matching the output of layer.get_weights()
       sess: tfe session"""
 
-    # Define keras weights as private variable
-    tfe_weights_pl = [tfe.define_private_placeholder(w.shape)
-                      for w in weights]
-
     # Assign new keras weights to existing weights defined by
     # default when tfe layer was instantiated
     if not sess:
       sess = K.get_session()
     for i, w in enumerate(self.weights):
-      fd = tfe_weights_pl[i].feed(weights[i])
-      sess.run(tfe.assign(w, tfe_weights_pl[i]), feed_dict=fd)
+      shape = w.shape.as_list()
+      tfe_weights_pl = tfe.define_private_placeholder(shape)
+      fd = tfe_weights_pl.feed(weights[i].reshape(shape))
+      sess.run(tfe.assign(w, tfe_weights_pl), feed_dict=fd)
 
   @property
   def prot(self):
--- a/tf_encrypted/keras/layers/convolutional.py
+++ b/tf_encrypted/keras/layers/convolutional.py
@@ -7,7 +7,6 @@ from tensorflow.python.keras.utils import conv_utils
 
 from tf_encrypted.keras.engine import Layer
 from tf_encrypted.keras import activations
-
 from tf_encrypted.keras.layers.layers_utils import not_implemented_arg_err
 
 logger = logging.getLogger('tf_encrypted')
@@ -145,7 +144,9 @@ class Conv2D(Layer):
     self.kernel = self.add_weight(kernel)
 
     if self.use_bias:
-      bias_shape = self.compute_output_shape(input_shape)[1:]
+      # Expand bias shape, will be broadcasted when
+      # added to the output
+      bias_shape = [self.filters, 1, 1]
       bias = self.bias_initializer(bias_shape)
       self.bias = self.add_weight(bias)
     else:
--- a/tf_encrypted/keras/layers/convolutional.py
+++ b/tf_encrypted/keras/layers/convolutional.py
@@ -7,7 +7,7 @@ from tensorflow.python.keras.utils import conv_utils
 
 from tf_encrypted.keras.engine import Layer
 from tf_encrypted.keras import activations
-from tf_encrypted.keras.layers.layers_utils import not_implemented_arg_err
+from tf_encrypted.keras.layers.layers_utils import default_args_check
 
 logger = logging.getLogger('tf_encrypted')
 
@@ -120,14 +120,14 @@ class Conv2D(Layer):
     self.bias_initializer = initializers.get(bias_initializer)
 
     # Not implemented arguments
-    not_implemented_arg_err(dilation_rate, "dilation_rate", "Conv2D")
-    not_implemented_arg_err(kernel_regularizer, "kernel_regularizer", "Conv2D")
-    not_implemented_arg_err(bias_regularizer, "bias_regularizer", "Conv2D")
-    not_implemented_arg_err(activity_regularizer,
-                            "activity_regularizer",
-                            "Conv2D")
-    not_implemented_arg_err(kernel_constraint, "kernel_constraint", "Conv2D")
-    not_implemented_arg_err(bias_constraint, "bias_constraint", "Conv2D")
+    default_args_check(dilation_rate, "dilation_rate", "Conv2D")
+    default_args_check(kernel_regularizer, "kernel_regularizer", "Conv2D")
+    default_args_check(bias_regularizer, "bias_regularizer", "Conv2D")
+    default_args_check(activity_regularizer,
+                       "activity_regularizer",
+                       "Conv2D")
+    default_args_check(kernel_constraint, "kernel_constraint", "Conv2D")
+    default_args_check(bias_constraint, "bias_constraint", "Conv2D")
 
   def build(self, input_shape):
     if self.data_format == 'channels_first':
@@ -144,8 +144,8 @@ class Conv2D(Layer):
     self.kernel = self.add_weight(kernel)
 
     if self.use_bias:
-      # Expand bias shape, will be broadcasted when
-      # added to the output
+      # Expand bias shape dimensions. Bias needs to have
+      # a rank of 3 to be added to the output
       bias_shape = [self.filters, 1, 1]
       bias = self.bias_initializer(bias_shape)
       self.bias = self.add_weight(bias)
--- a/tf_encrypted/keras/layers/convolutional.py
+++ b/tf_encrypted/keras/layers/convolutional.py
@@ -8,7 +8,8 @@ from tensorflow.python.keras.utils import conv_utils
 from tf_encrypted.keras.engine import Layer
 from tf_encrypted.keras import activations
 
-arg_not_impl_msg = "`{}` argument is not implemented for layer {}"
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

File: tf_encrypted/keras/layers/layers_utils.py [hunk 1]
```diff
@@ -0,0 +1,25 @@
+"""TF Encrypted Keras layers utils"""
+import inspect
+
+import tensorflow as tf
+
+
+def not_implemented_arg_err(arg, arg_name, tf_layer_name):
```
jvmancuso: I think this should be implemented as a subclass of `NotImplementedError` and renamed to something like `LayerArgNotImplementedError`.  The arguments to this function would become arguments to the subclass's constructor, and you can override the `__str__` method of the subclass to return a custom error message.  The logic of this function can be implemented in the constructor or in the `__str__` method, whichever seems best.

yanndupis: Sounds good, let's do that.

yanndupis: I have subclassed `NotImplementedError`. However not sure how the following logic could be included in the class and how `LayerArgNotImplementedError` would be called in each layer:

```
tf_layer_cls = getattr(tf.keras.layers, tf_layer_name)
layer_sign = inspect.signature(tf_layer_cls.__init__).parameters
 tf_default_arg = layer_sign[arg_name].default
 if arg != tf_default_arg:
   raise LayerArgNotImplementedError(arg_name, tf_layer_name, tf_default_arg)
```

That's why I kept for now `not_implemented_arg_err`.

tf_encrypted/keras/engine/base_layer.py
```diff
@@ -115,17 +115,15 @@ def set_weights(self, weights, sess=None):
...
     # default when tfe layer was instantiated
     if not sess:
       sess = K.get_session()
     for i, w in enumerate(self.weights):
-      fd = tfe_weights_pl[i].feed(weights[i])
-      sess.run(tfe.assign(w, tfe_weights_pl[i]), feed_dict=fd)
+      shape = w.shape.as_list()
+      tfe_weights_pl = tfe.define_private_placeholder(shape)
+      fd = tfe_weights_pl.feed(weights[i].reshape(shape))
+      sess.run(tfe.assign(w, tfe_weights_pl), feed_dict=fd)
```
yanndupis: I had to change a bit the logic here so the placeholder shape gets defined based on the layer shape weights instead of numpy arrays from `get_weights`. It was causing some issue for the bias in the conv2d. The shape of the numpy array is rank 1 [filter], when the conv2d layer needs a bias of rank=3

jvmancuso: I'm not sure if this helps, but I had imagined this being two separate sets of logic -- one for numpy arrays and one for TF Variables.

yanndupis:
Hey @jvmancuso, I think it's ready for another review before being merged.

tf_encrypted/keras/layers/convolutional.py
```diff
@@ -160,7 +144,9 @@ def build(self, input_shape):
     self.kernel = self.add_weight(kernel)
 
     if self.use_bias:
-      bias_shape = self.compute_output_shape(input_shape)[1:]
+      # Expand bias shape dimensions. Bias needs to have
+      # a rank of 3 to be added to the output
+      bias_shape = [self.filters, 1, 1]
```
jvmancuso: this seems weird -- it should be able to broadcast. okay with this code but it smells like a bug elsewhere

tf_encrypted/keras/layers/convolutional_test.py
```diff
@@ -16,7 +16,10 @@ def setUp(self):
     tf.reset_default_graph()
 
   def test_conv2d_bias(self):
-    self._core_conv2d(kernel_size=2, use_bias=True)
+    self._core_conv2d(kernel_size=2,
+                      use_bias=True,
+                      dilation_rate=(1, 1),
```
jvmancuso: are new args necessary? thought they were the default

tf_encrypted/keras/layers/layers_utils.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class LayerArgNotImplementedError(NotImplementedError):
+  """Raise error when layer argument is not yet supported in TFE
+
+    Args:
+      arg: TFE layer argument
+      arg_name: TFE layer argument name (string)
+      tf_layer_name: Tensorflow keras layer name (string)
+
```
jvmancuso: there's an extra line here

tf_encrypted/keras/layers/normalization.py
```diff
@@ -112,46 +113,36 @@ def __init__(self,
...
+    default_args_check(renorm,
+                       "renorm",
+                       "BatchNormalization")
+    default_args_check(renorm_clipping,
+                       "renorm_clipping",
+                       "BatchNormalization")
+    default_args_check(fused,
+                       "fused",
+                       "BatchNormalization")
+    default_args_check(trainable,
```
jvmancuso: I don't think we should be doing this for trainable -- we don't support training so this kwarg is meaningless for the tfe layer, so we should support either value

tf_encrypted/keras/layers/normalization.py
```diff
@@ -112,46 +113,36 @@ def __init__(self,
...
+                       "gamma_constraint",
+                       "BatchNormalization")
+    default_args_check(renorm,
+                       "renorm",
+                       "BatchNormalization")
+    default_args_check(renorm_clipping,
+                       "renorm_clipping",
+                       "BatchNormalization")
+    default_args_check(fused,
+                       "fused",
```
jvmancuso: same comment as for `trainable`

tf_encrypted/keras/models/sequential_test.py
```diff
@@ -87,6 +88,41 @@ def test_clone_model(self):
...
+  def test_conv_model(self):
+
+    num_classes = 10
+    input_shape = (1, 28, 28, 1)
+    input_data = np.random.normal(size=input_shape)
+
+    with tf.Session():
+      model = tf.keras.models.Sequential()
+
+      model.add(tf.keras.layers.Conv2D(10,
```
jvmancuso: this value of channels seems high for a test -- can't we test the same with just 2?

tf_encrypted/keras/models/sequential_test.py
```diff
@@ -87,6 +88,41 @@ def test_clone_model(self):
...
+                                       (3, 3),
+                                       batch_input_shape=input_shape))
+      model.add(tf.keras.layers.AveragePooling2D((2, 2)))
+      model.add(tf.keras.layers.Conv2D(32, (3, 3)))
```
jvmancuso: same comment as above

tf_encrypted/keras/testing_utils.py
```diff
@@ -26,6 +26,13 @@ def agreement_test(tfe_layer_cls, kwargs=None, input_shape=None,
   tf_layer_cls = getattr(tf.keras.layers, tfe_layer_cls.__name__)
   tfe_kwargs = {**kwargs, **tfe_kwargs}
 
+  # If the layer as a `fused` argument (e.g. BatchNormalization),
+  # the TF layer should be executed with fused=False. The main reason is
+  # that TF supports channel first only when fused=False
+  if 'fused' in list(kwargs):
```
jvmancuso: I think you can remove the `list` and just use `if `fused` in kwargs:`

tf_encrypted/keras/testing_utils.py
```diff
@@ -26,6 +26,13 @@ def agreement_test(tfe_layer_cls, kwargs=None, input_shape=None,
   tf_layer_cls = getattr(tf.keras.layers, tfe_layer_cls.__name__)
   tfe_kwargs = {**kwargs, **tfe_kwargs}
 
+  # If the layer as a `fused` argument (e.g. BatchNormalization),
+  # the TF layer should be executed with fused=False. The main reason is
+  # that TF supports channel first only when fused=False
```
jvmancuso: I don't think this should matter since our implementation can handle both, right?

tf_encrypted/keras/layers/convolutional_test.py
yanndupis: they are not, I will remove it. thanks

tf_encrypted/keras/layers/normalization.py
yanndupis: Agreed, i will remove it.

tf_encrypted/keras/layers/normalization.py
yanndupis: agreed

tf_encrypted/keras/testing_utils.py
yanndupis: I was getting a pb because `default_args_check(fused,..,..)` was expecting to have `fused=None` but it had to be `fused=False`  when running the test, otherwise TensorFlow couldn't run with `fused=None` and channel_first. 

But I am going to remove this meaningless test `default_args_check(fused,..,..)` so won't need to do this check.

tf_encrypted/keras/models/sequential_test.py
yanndupis: good point, we should reduce it.

PR #590, opened 2019-06-27
Create tfe.Session() when setting weights

yanndupis:
Hey,

Until now we had to `set_weights` under `with tf.Session() as sess:` or pass explicitly a TFE session. However ideally we would like to be able to `set_weights` outside this context to get the same behavior as Keras:

```
model = tf.keras.models.Sequential()
model.add(tf.keras.layers.Dense(5, use_bias=False, input_shape=[5]))
model.set_weights([np.ones([5, 5])])
y = model(x)

with K.get_session() as sess:
    sess.run(y)
```

This PR is a potential solution but it's not perfect. Ideally we would like to have something like `K.get_session()` which would create a tfe.Session() when it doesn't exist. Currently to have tfe.Session() from `K.get_session()`, it has to be previously set with `K.set_session(sess)`.

Thanks,

### DIFF
--- a/tf_encrypted/keras/__init__.py
+++ b/tf_encrypted/keras/__init__.py
@@ -5,6 +5,7 @@ from tf_encrypted.keras import engine
 from tf_encrypted.keras import models
 from tf_encrypted.keras import layers
 from tf_encrypted.keras.models import Sequential
+from tf_encrypted.keras import backend
 
 
 __all__ = [
@@ -12,4 +13,5 @@ __all__ = [
     'models',
     'layers',
     'Sequential',
+    'backend'
 ]
--- a/tf_encrypted/keras/__init__.py
+++ b/tf_encrypted/keras/__init__.py
@@ -1,17 +1,17 @@
 """Higher-level layer abstractions built on TF Encrypted."""
 from __future__ import absolute_import
 
-from tf_encrypted.keras import engine
-from tf_encrypted.keras import models
-from tf_encrypted.keras import layers
-from tf_encrypted.keras.models import Sequential
 from tf_encrypted.keras import backend
+from tf_encrypted.keras import engine
+from tf_encrypted.keras import layers
+from tf_encrypted.keras import models
+from tf_encrypted.keras.models import Sequential
 
 
 __all__ = [
+    'backend',
     'engine',
-    'models',
     'layers',
+    'models',
     'Sequential',
-    'backend'
 ]
--- a/tf_encrypted/keras/backend.py
+++ b/tf_encrypted/keras/backend.py
@@ -1,17 +1,7 @@
 """TFE Keras backend"""
-import tensorflow as tf
 from tensorflow.keras import backend as K
 
 import tf_encrypted as tfe
-from tf_encrypted.protocol.pond import PondPrivateVariable
-
-
-# TODO Should be moved outside backend, maybe in pond.py?
-def is_variable_initialized(x):
-  """Test if TFE private variable has been initialized"""
-  assert isinstance(x, PondPrivateVariable), type(x)
-  is_initialized = tf.is_variable_initialized(x.share0.variable)
-  return is_initialized
 
 
 def get_session():
--- a/tf_encrypted/keras/backend.py
+++ b/tf_encrypted/keras/backend.py
@@ -1,9 +1,11 @@
-"""TFE Keras backend"""
+"""TFE Keras backend. Most of the code was
+borrowed from the tf.keras codebase.
+"""
 import threading
 
 from tensorflow.python.framework import ops
-from tensorflow.python.ops import array_ops
 from tensorflow.python.keras.backend import get_graph, name_scope
+from tensorflow.python.ops import array_ops
 
 import tf_encrypted as tfe
 
@@ -20,7 +22,8 @@ def get_session(op_input_list=()):
     session = default_session
     # If the default session is not a TFE Session, create one
     if not isinstance(session, tfe.Session()):
-      session = tfe.Session()
+      _SESSION.session = tfe.Session()
+      session = _SESSION.session
   else:
     if ops.inside_function():
       raise RuntimeError('Cannot get session inside Tensorflow graph function.')
--- a/tf_encrypted/keras/backend.py
+++ b/tf_encrypted/keras/backend.py
@@ -22,7 +22,7 @@ def get_session(op_input_list=()):
     # If the default session is a TFE Session return this session
     if isinstance(default_session, tfe.Session()):
       return default_session
-    else:
+    if not isinstance(default_session, tfe.Session()):
       raise TypeError(
           'The default session should be a tfe.Session(). '
           'You are probably trying to run this graph with '
@@ -35,7 +35,8 @@ def get_session(op_input_list=()):
     if (getattr(_SESSION, 'session', None) is None or
         _SESSION.session.graph is not _current_graph(op_input_list)):
       _SESSION.session = tfe.Session()
-      return _SESSION.session
+      session = _SESSION.session
+    return session
 
 
 def _current_graph(op_input_list):
--- a/tf_encrypted/keras/backend.py
+++ b/tf_encrypted/keras/backend.py
@@ -1,0 +1,34 @@
+"""TFE Keras backend"""
+import tensorflow as tf
+from tensorflow.keras import backend as K
+
+import tf_encrypted as tfe
+from tf_encrypted.protocol.pond import PondPrivateVariable
+
+
+# TODO Should be moved outside backend, maybe in pond.py?
+def is_variable_initialized(x):
+  """Test if TFE private variable has been initialized"""
+  assert isinstance(x, PondPrivateVariable), type(x)
+  is_initialized = tf.is_variable_initialized(x.share0.variable)
+  return is_initialized
+
+
+def get_session():
+  """Returns the TFE session to be used"""
+  sess = K.get_session()
+  # If session is not a tfe.Session, creates one
+  # K.get_session() will return a tfe.Session if
+  # TFE global session was previously set with set_session
+  if not isinstance(sess, tfe.Session):
+    sess = tfe.Session()
+
+  return sess
+
+def set_session(sess):
+  """Sets the global TFE session."""
+  return K.set_session(sess)
+
+def clear_session():
+  """Destroys the current TFE graph and creates a new one"""
+  return K.clear_session()
--- a/tf_encrypted/keras/backend.py
+++ b/tf_encrypted/keras/backend.py
@@ -35,8 +35,8 @@ def get_session(op_input_list=()):
     if (getattr(_SESSION, 'session', None) is None or
         _SESSION.session.graph is not _current_graph(op_input_list)):
       _SESSION.session = tfe.Session()
-      session = _SESSION.session
-    return session
+    session = _SESSION.session
+  return session
 
 
 def _current_graph(op_input_list):
--- a/tf_encrypted/keras/backend.py
+++ b/tf_encrypted/keras/backend.py
@@ -4,8 +4,6 @@ borrowed from the tf.keras codebase.
 import threading
 
 from tensorflow.python.framework import ops
-from tensorflow.python.keras.backend import get_graph, name_scope
-from tensorflow.python.ops import array_ops
 
 import tf_encrypted as tfe
 
@@ -17,13 +15,18 @@ _SESSION = threading.local()
 def get_session(op_input_list=()):
   """Returns the session object for the current thread."""
   global _SESSION
+  if getattr(_SESSION, 'session', None) is not None:
+    return _SESSION.session
   default_session = ops.get_default_session()
   if default_session is not None:
-    session = default_session
-    # If the default session is not a TFE Session, create one
-    if not isinstance(session, tfe.Session()):
-      _SESSION.session = tfe.Session()
-      session = _SESSION.session
+    # If the default session is a TFE Session return this session
+    if isinstance(default_session, tfe.Session()):
+      return default_session
+    else:
+      raise TypeError(
+          'The default session should be a tfe.Session(). '
+          'You are probably trying to run this graph with '
+          'tf.Session() instead of tfe.Session()')
   else:
     if ops.inside_function():
       raise RuntimeError('Cannot get session inside Tensorflow graph function.')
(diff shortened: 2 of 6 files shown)


jvmancuso:
I have a suggestion.  What if we create a `tfe.keras.backend` module that provides a thin wrapper for the default tf.Session that converts to and from tfe.Session?  Then we can use:

```python
from tf_encrypted.keras import backend as KE

model = tf.keras.models.Sequential()
model.add(tf.keras.layers.Dense(5, use_bias=False, input_shape=[5]))
model.set_weights([np.ones([5, 5])])
y = model(x)

with KE.get_session() as sess:  # takes the current sess living in tf.keras.backend,
    sess.run(y)                 # wraps it in a tfe.Session, and returns it

```

To do this, I think we'd want to implement a classmethod constructor, e.g. `tfe.Session.from_vanilla_session`, with usage like: `sess = tfe.Session.from_vanilla_session(tf.Session(), **tfe_session_kwargs)`

Then `tfe.keras.backend.get_session()` would be defined this way:

```python
# Lives in tf_encrypted/keras/backend.py
from tensorflow.keras import backend as K

def get_session():
  return tfe.Session.from_vanilla_session(K.get_session())
```

Similarly, we could implement our own version of `set_session()`.  I think this would make our code more straightforward, so we can stick with a Keras way of managing sessions that's TFE-aware

@mortendahl @yanndupis what do you think?

yanndupis:
I like this idea :).

tf_encrypted/keras/models/sequential.py
```diff
@@ -112,15 +113,21 @@ def set_weights(self, weights, sess=None):
           matching the output of model.get_weights()
       sess: tfe session"""
 
+    if not sess:
+      sess = tfe.Session()
+      # Set as keras global session so the
+      # model can be run with K.get_session():
+      K.set_session(sess)
+      # Initialize model weights before setting weights
+      sess.run(tf.global_variables_initializer())
```
mortendahl: When/if we move to a `set_session` approach then all `sess.run(tf.global_variables_initializer())` are dangerous, as they will potentially override existing values.

Two things to keep in mind:

1. Variables that have been explicitly assigned to do not need to initialised AFAICT
2. It might be better to keep track of each variable used in TFE Keras (this is done in TF Keras as well) and whether or not it has been initialised; setting the variable should also mark it as initialised, per 1.

In this approach variables are initialised when the model is first used.

tf_encrypted/keras/models/sequential_test.py
```diff
@@ -53,14 +53,14 @@ def test_from_config(self):
...
 
     with tfe.protocol.SecureNN():
-      tfe_model = tfe.keras.models.Sequential([])
-      tfe_model = tfe_model.from_config(k_config)
       x = tfe.define_private_variable(input_data)
 
-    with tfe.Session() as sess:
-      sess.run(tf.global_variables_initializer())
+      tfe_model = tfe.keras.models.Sequential([])
+      tfe_model = tfe_model.from_config(k_config)
```
mortendahl: this is a bit of a strange pattern (instantiate a model just to throw it away); is there an equivalent `from_config` in TF?

jvmancuso: I think this is just improper usage of the existing method, since it's a classmethod and doesn't require the instantiation here

yanndupis: ah correct, the following would work: 
```
tfe_model = Sequential.from_config(k_config)
tfe_model.set_weights(k_weights)
```

tf_encrypted/keras/models/sequential.py
yanndupis: ah you are right. I ran some quick tests. If a variable has been explicitly assigned we don't need to initialize it.

yanndupis: I made couple updates:
- added get_session and set_session. For get session maybe there is a better approach other than checking if K.get_session() returns a tfe.Session
- When passing to set_weights a list of private variables, I checked if they have been initialized with `is_variable_initialized`. If not, we initialize them.  I am doing that so in [`test_weights_as_private_var`](https://github.com/tf-encrypted/tf-encrypted/blob/756547ac32473aea982b51e6c1899ee534e1725a/tf_encrypted/keras/models/sequential_test.py#L99) we don't have to manually create a tfe.Session to initialize the new weights then pass around the tfe.Session to `set_weights` etc.
- Note: model variables get tracked in [layer weights attribute](https://github.com/tf-encrypted/tf-encrypted/blob/756547ac32473aea982b51e6c1899ee534e1725a/tf_encrypted/keras/engine/base_layer.py#L111). I am not tracking if they have been initialized or not because variables that have been explicitly assigned to do not need to be initialized. If we need to check if the variable has been initialized, we could use is_variable_initialized`.

tf_encrypted/keras/backend.py
```diff
@@ -0,0 +1,34 @@
...
+
+import tf_encrypted as tfe
+from tf_encrypted.protocol.pond import PondPrivateVariable
+
+
+# TODO Should be moved outside backend, maybe in pond.py?
+def is_variable_initialized(x):
+  """Test if TFE private variable has been initialized"""
+  assert isinstance(x, PondPrivateVariable), type(x)
+  is_initialized = tf.is_variable_initialized(x.share0.variable)
```
mortendahl: don't think it's a good idea to break encapsulation here. a better approach would be to move this logic to the variables themselves and simply call `x.is_initialized`

tf_encrypted/keras/backend.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def get_session():
+  """Returns the TFE session to be used"""
+  sess = K.get_session()
+  # If session is not a tfe.Session, creates one
+  # K.get_session() will return a tfe.Session if
+  # TFE global session was previously set with set_session
+  if not isinstance(sess, tfe.Session):
+    sess = tfe.Session()
```
mortendahl: each consecutive `get_session` will create a new `tfe.Session`, not sure that's what we want. i'm not convinced we should mix, or try to reuse, the `tf.keras` session but rather manage our own. thoughts? cc @yanndupis @jvmancuso

tf_encrypted/keras/engine/base_layer.py
```diff
@@ -135,6 +136,8 @@ def set_weights(self, weights, sess=None):
         sess.run(tfe.assign(w, tfe_weights_pl), feed_dict=fd)
     elif isinstance(weights[0], PondPrivateTensor):
       for i, w in enumerate(self.weights):
+        if not sess.run(KE.is_variable_initialized(weights[i])):
```
mortendahl: might be worth changing the order of `if isinstance` and `for i, w`, should allow the user to provide a mix of different types

tf_encrypted/keras/engine/base_layer.py
```diff
@@ -135,6 +136,8 @@ def set_weights(self, weights, sess=None):
...
+          sess.run(weights[i].initializer)
```
mortendahl: you're only testing that `weights[i]` are private tensors so not guaranteed to have an initializer (only variables do). i'm not sure any initializers should be called here; if the user passes in a private variable then it's his responsibility to ensure that it's already initialized.

what does TF Keras do if a user passes in a variable?

tf_encrypted/keras/models/sequential.py
```diff
@@ -112,15 +111,21 @@ def set_weights(self, weights, sess=None):
           matching the output of model.get_weights()
       sess: tfe session"""
 
+    if not sess:
+      #sess = tfe.Session()
+      sess = KE.get_session()
+      # Set as keras global session so the
+      # model can be run with K.get_session():
+      sess = KE.get_session()
+      KE.set_session(sess)
```
mortendahl: this should happen in `get_session`, if anywhere

tf_encrypted/keras/backend.py
jvmancuso: I agree.  See how tf.keras does it [here](https://github.com/tensorflow/tensorflow/blob/88340a942c897601cc4d31c7c61bf68f60446857/tensorflow/python/keras/backend.py#L91) and [here](https://github.com/tensorflow/tensorflow/blob/88340a942c897601cc4d31c7c61bf68f60446857/tensorflow/python/keras/backend.py#L422)

tf_encrypted/keras/engine/base_layer.py
jvmancuso: users are not allowed to do so in tf.keras -- they restrict to numpy arrays, but that's not enough for us since we want to be able to do so without reconstructing the weights on the device of whoever runs this code

jvmancuso: agreed about initializers though, let's assume the users have run these on their own whenever needed

yanndupis: Sounds good, let's assume, they have been run.

yanndupis:
Hey guys, I implemented a custom get_session, set_session etc. with tfe.Session. Should be better now. Thanks in advance for your review.

tf_encrypted/keras/__init__.py
```diff
@@ -5,11 +5,13 @@
...
 from tf_encrypted.keras.models import Sequential
+from tf_encrypted.keras import backend
 
 
 __all__ = [
     'engine',
     'models',
     'layers',
     'Sequential',
+    'backend'
```
mortendahl: list not sorted; make sure each line has a `,` at the end (this leaves all existing lines unchanged when more are added)

tf_encrypted/keras/backend.py
```diff
@@ -0,0 +1,59 @@
+"""TFE Keras backend"""
```
mortendahl: some of this code seems to come from TF Keras, in which case we should at least mention that in the file header

tf_encrypted/keras/engine/base_layer.py
```diff
@@ -3,10 +3,10 @@
 import logging
 
 import numpy as np
-from tensorflow.keras import backend as K
 from tensorflow.python.keras.utils import generic_utils
 
 import tf_encrypted as tfe
+from tf_encrypted.keras import backend as KE
```
mortendahl: imports not sorted

tf_encrypted/keras/backend.py
```diff
@@ -0,0 +1,59 @@
...
+
+def get_session(op_input_list=()):
+  """Returns the session object for the current thread."""
+  global _SESSION
+  default_session = ops.get_default_session()
+  if default_session is not None:
+    session = default_session
+    # If the default session is not a TFE Session, create one
+    if not isinstance(session, tfe.Session()):
+      session = tfe.Session()
```
mortendahl: if the session is created here it won't be reused since it's not stored back onto `_SESSION`, don't think this is intended behaviour

tf_encrypted/keras/models/sequential.py
```diff
@@ -112,15 +111,20 @@ def set_weights(self, weights, sess=None):
           matching the output of model.get_weights()
       sess: tfe session"""
 
+    if not sess:
+      #sess = tfe.Session()
+      sess = KE.get_session()
+      # Set as keras global session so the
+      # model can be run with K.get_session():
+      sess = KE.get_session()
```
mortendahl: leftover?

tf_encrypted/protocol/pond/__init__.py
```diff
@@ -2,7 +2,7 @@
 from __future__ import absolute_import
 
 from .pond import Pond
-from .pond import PondTensor, PondPublicTensor, PondPrivateTensor, PondMaskedTensor
+from .pond import PondTensor, PondPublicTensor, PondPrivateTensor, PondMaskedTensor, PondPrivateVariable
```
mortendahl: maybe change to single import per line

tf_encrypted/keras/backend.py
yanndupis: ah good catch!

tf_encrypted/keras/models/sequential.py
yanndupis: oups, yes. thanks

tf_encrypted/keras/backend.py
yanndupis: good point

yanndupis:
Thanks for your review @mortendahl, should be good now.

tf_encrypted/keras/backend.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def clear_session():
+  """Destroys the current TFE graph and creates a new one"""
+  _SESSION.session = None
+  ops.reset_default_graph()
+  graph = get_graph()
+  with graph.as_default():
+    with name_scope(''):
+      array_ops.placeholder_with_default(
```
mortendahl: what is this used for?

tf_encrypted/keras/backend.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def get_session(op_input_list=()):
+  """Returns the session object for the current thread."""
+  global _SESSION
+  default_session = ops.get_default_session()
+  if default_session is not None:
+    session = default_session
+    # If the default session is not a TFE Session, create one
+    if not isinstance(session, tfe.Session()):
```
mortendahl: i think there's still a flaw in the logic here. if `default_session` is a TF session, then any TFE session already in `_SESSION` will be ignored and a new one created.

maybe something like:
1. if `_SESSION.session` is not none then return that
2. if `ops.get_default_session` is a TFE session then return that
3. create a new TFE session, store in `_SESSION.session`, and return that

yanndupis: my bad, apologize, tried to fix this too quickly yesterday. good catch. thanks

tf_encrypted/keras/backend.py
yanndupis: ah in fact, we don't need this. we just need to reset de graph and set global session `_SESSION` to None.

https://github.com/tensorflow/tensorflow/blob/c407b045b8802f9eded430ef48be18cd85e4788c/tensorflow/python/keras/backend.py#L218

tf_encrypted/keras/backend.py
```diff
@@ -0,0 +1,59 @@
...
+  global _SESSION
+  if getattr(_SESSION, 'session', None) is not None:
+    return _SESSION.session
+  default_session = ops.get_default_session()
+  if default_session is not None:
+    # If the default session is a TFE Session return this session
+    if isinstance(default_session, tfe.Session()):
+      return default_session
+    if not isinstance(default_session, tfe.Session()):
+      raise TypeError(
```
mortendahl: what's the reason for this? it seems that in most situations the user would have to manually set a session first for tfe.keras to work -- maybe this is expected behaviour..?

yanndupis: I have added this error in case, the model definition or set_weights is ran under `tf.Session()` instead of tfe.Session().

PR #597, opened 2019-07-04
Only show Keras API docs

bendecoste:
Closes https://github.com/tf-encrypted/tf-encrypted/issues/587

Other docs are out of date right now.
This PR introduces the follwing:

* Started using [naploean](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html).
  In short, napolean generates restructured text from google-style docstrings.
* Only show the keras API

There is one thing that I am not happy with, and that is that the titles generated
by sphinx-apidoc are bad.  E.g. for the `tf_encrypted.keras` page, the page title will be
`tf_encrypted.keras package`.  Where something like `Keras` or `TFE Keras` would be preferable.

Apparently using `-M` in the apidoc command should use the module title instead, but this does
not appear to be happening.

https://github.com/sphinx-doc/sphinx/pull/6492/ has introduced a `--template` command, which
we should be able to eventually leverage to fix the above issue.  Whether we want to wait or
take a different approach with this is up to us.  I can also manually edit the .rst files
and change the header if we want.  I just wanted to get a discussion going.


bendecoste:
If we know which docs are up to date I can keep them around.  Merging this would make it so ONLY keras docs are shown.

docs/source/index.rst
```diff
@@ -16,12 +16,12 @@ In particular, the library focuses on:
...
 
-Checkout the `Getting Started`_ guide to learn how to get up and running with private machine learning.
+This page only contains API documentation. Checkout the `examples`_ on github to learn how to get up and running with private machine learning.
 
 You can view the project source, contribute, and asks questions on `GitHub`_.
 
 .. _TensorFlow: https://www.tensorflow.org
-.. _Getting Started: usage/getting_started.html
+.. _examples: https://github.com/tf-encrypted/tf-encrypted/tree/master/examples
 .. _GitHub: https://github.com/mortendahl/tf-encrypted
```
mortendahl: old URL

docs/source/index.rst
```diff
@@ -3,7 +3,7 @@
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
 
-TF Encrypted
+TF Encrypted API Docs
 ========================================
 
 TF Encrypted is a Python library built on top of TensorFlow_ for researchers and practitioners to experiment with privacy-preserving machine learning. It provides an interface similar to that of TensorFlow, and aims at making the technology readily available without first becoming an expert in machine learning, cryptography, distributed systems, and high performance computing.
```
mortendahl: this text could be updated, eg from the latest readme

bendecoste:
> lgtm, only real question is around the files in `gen`: are they autogenerated? and if so, is it worth having them in the repo?

We can gitignore them, if we wanted to manually rename the headers we would need to keep them.

mortendahl:
ready to merge this @bendecoste?

bendecoste:
Yes if you are!

PR #598, opened 2019-07-04
Local computation decorator

jvmancuso:
Adds a `local_computation` decorator that wraps up functions in a call to `define_local_computation`. See examples for usage.

Other small fixes:
- Small style fixes
- Changed the way metrics are reported in the secure aggregation example.
- For reconstruct_input in tfe.define_local_computation, use tf.contrib.framework.nest when dealing with iterables of tensors. Planning to open an issue to standardize this pattern across the project.

### DIFF
--- a/examples/federated-learning/run.py
+++ b/examples/federated-learning/run.py
@@ -92,6 +92,7 @@ class ModelOwner:
     iterator = dataset.make_one_shot_iterator()
     return iterator.get_next()
 
+  @tfe.local_computation
   def update_model(self, *grads):
     """Perform a single update step.
 
@@ -108,14 +109,16 @@ class ModelOwner:
           param.assign(param - grad * self.LEARNING_RATE)
           for param, grad in zip(params, grads)
       ])
-      # return update_op
 
     with tf.name_scope('validate'):
       x, y = self._build_data_pipeline()
       y_hat, loss = self._build_validation_step(x, y)
 
       with tf.control_dependencies([update_op]):
-        return tf.print('expect', loss, y, y_hat, summarize=50)
+        print_loss = tf.print("loss", loss)
+        print_expected = tf.print("expect", y, summarize=50)
+        print_result = tf.print("result", y_hat, summarize=50)
+        return print_loss, print_expected, print_result
 
 
 class DataOwner:
@@ -150,6 +153,7 @@ class DataOwner:
     iterator = dataset.make_one_shot_iterator()
     return iterator.get_next()
 
+  @tfe.local_computation
   def compute_gradient(self):
     """Compute gradient given current model parameters and local data."""
     with tf.name_scope('data_loading'):
@@ -174,8 +178,7 @@ if __name__ == "__main__":
   ]
 
   model_grads = zip(*(
-      tfe.define_private_input(data_owner.player_name,
-                               data_owner.compute_gradient)
+      data_owner.compute_gradient()
       for data_owner in data_owners
   ))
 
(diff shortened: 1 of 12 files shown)


examples/federated-learning/run.py
```diff
@@ -108,14 +109,16 @@ def update_model(self, *grads):
...
     with tf.name_scope('validate'):
       x, y = self._build_data_pipeline()
       y_hat, loss = self._build_validation_step(x, y)
 
       with tf.control_dependencies([update_op]):
-        return tf.print('expect', loss, y, y_hat, summarize=50)
+        print_loss = tf.print("loss", loss)
+        print_expected = tf.print("expect", y, summarize=50)
+        print_result = tf.print("result", y_hat, summarize=50)
+        return print_loss, print_expected, print_result
```
mortendahl: you should probably group these guys (the internal grouping is more a security check that we should not rely on)

examples/logistic/prediction_joint.py
```diff
@@ -10,17 +10,14 @@
 prediction_client_1 = PredictionClient('prediction-client-1', num_features // 2)
 result_receiver = prediction_client_0
 
-x_0 = tfe.define_private_input(prediction_client_0.player_name,
-                               prediction_client_0.provide_input)
-x_1 = tfe.define_private_input(prediction_client_1.player_name,
-                               prediction_client_1.provide_input)
+x_0 = prediction_client_0.provide_input()
+x_1 = prediction_client_0.provide_input()
```
mortendahl: should be `prediction_client_1`

mortendahl: since there's a bit more magic now, maybe it's worth adding some assertions to this example, just to show the user that eg `x_0` and `x_1` are private tensors. once we refactor we can also assert something like `x_0.owner == prediction_client_0.player`

examples/simple-average/run.py
```diff
@@ -15,31 +14,27 @@
...
 if __name__ == '__main__':
   # get input from inputters as private values
-  inputs = [
-      tfe.define_private_input('inputter-0', provide_input),
-      tfe.define_private_input('inputter-1', provide_input),
-      tfe.define_private_input('inputter-2', provide_input),
-      tfe.define_private_input('inputter-3', provide_input),
-      tfe.define_private_input('inputter-4', provide_input),
-  ]
+  inputs = [provide_input(player_name="inputter-{}".format(i))  # pylint: disable=unexpected-keyword-arg
```
mortendahl: fyi the explicit list was intentional to make it super clear that we're dealing with five inputters

examples/simple-average/run.py
```diff
@@ -15,31 +14,27 @@
   tfe.set_config(config)
   tfe.set_protocol(tfe.protocol.Pond())
 
-
+@tfe.local_computation
 def provide_input() -> tf.Tensor:
   # pick random tensor to be averaged
   return tf.random_normal(shape=(10,))
 
+@tfe.local_computation('result-receiver')
```
mortendahl: maybe keep the same syntax as for `provide_input`, and give the player name when the function is called instead

tf_encrypted/protocol/pond/pond.py
```diff
@@ -439,6 +439,119 @@ def helper(v: tf.Tensor) -> "PondPublicTensor":
...
+          is_op = tf.contrib.framework.nest.flatten(
+              tf.contrib.framework.nest.map_structure(
+                  lambda x: isinstance(x, tf.Operation),
+                  output,
+              )
+          )
+          if all(is_op):
+            return tf.group(output)
+
+          return output
```
mortendahl: this bypasses the security precaution even if only a single value is not an op. until we need it, it might be better to simply assert that only `tf.Operation`s are returned, and always group

examples/simple-average/run.py
jvmancuso: I preferred this way because it demonstrates both ways of using it. Maybe it would be best to add a comment that emphasizes that, or we can do that in different examples (just not sure we should expect people to pick up details from all examples)

examples/logistic/prediction_joint.py
jvmancuso: Up to you!

Issue #460, opened 2019-04-30
Depthwise separable 2D convolution


avitalsh:
Hi,

Is there a plan to add support for depthwise separable 2D convolution (DepthwiseConv2D) ? 

Thanks

jvmancuso:
Yes, it's on our roadmap, although we've prioritized a few other items (in particular, more robust support for Keras layers and models).  @yanndupis or I have some other things on our roadmap ahead of that, but we hoped to get to it in the next couple weeks.

If you'd like to submit a PR adding it to the Pond protocol, we'd be more than happy to work with you to merge it!

mortendahl:
@jvmancuso is this something we should bump up and look at within a week or two?

jvmancuso:
This would be great to work on after the keras work is done. The forward pass is easy, but the backward pass can be quite tricky.

avitalsh:
@jvmancuso Hey, is this planned to be added sometime soon?

jvmancuso:
We've had plans to do so, but it's been a bit neglected -- would you find it useful to have it?

If you'd like to add it, we'd be more than happy to help you with the implementation and get it merged.

PR #612, opened 2019-07-15
tfe.keras - depthwise conv2D

avitalsh:
Added support to depthwise convolution. 
Implementation is based on the paper **Diagonalwise Refactorization: An Efficient
Training Method for Depthwise Convolutions** (https://arxiv.org/pdf/1803.09926.pdf) and was expanded to support depth_multiplier > 1. 

Related to issue #460


avitalsh:
Thanks for the review @jvmancuso!
Regarding the gradient - I didn't really put much thought into this but the paper says that - 

> During backward propagation, the gradients from the
top layer are passed only to the diagonal weights and all other
positions remain 0.

and also - 
image

jvmancuso:
Yup, I'd seen the first sentence, just seemed somewhat ambiguous to me.  Anyway, that should be good enough to admit an implementation later on. LGTM!

PR #614, opened 2019-07-16
Added support for batches in secure_model

avitalsh:
Similar to the logic in bin/run


jvmancuso:
I was thinking that maybe there should be some caveat about warming-up the graph, but I think that's unrelated to this one, so good on my end!

PR #619, opened 2019-07-17
Added converter support for DepthwiseConv2D

avitalsh:
Added converter support the DepthwiseConv2D.
Two changes were made in converter.py - 
1. Added '^' to the regex in ```match_numbered_scope``` - this was needed because when looking for special ops, the ```depthwise_conv2d``` node matched to ```conv2d``` op. 

2. added '$' to the regex in ```match_numbered_leaf``` - this was needed because when looking for the interiors of the depthwise conv layer the ```depthwise``` op matched the ```depthwise_conv2d/depthwise_kernel``` node.

### DIFF
--- a/tf_encrypted/convert/register.py
+++ b/tf_encrypted/convert/register.py
@@ -10,7 +10,7 @@ import numpy as np
 import tensorflow as tf
 
 from ..layers import Conv2D, Relu, Sigmoid, Dense, AveragePooling2D, MaxPooling2D
-from ..keras.layers import BatchNormalization
+from ..keras.layers import BatchNormalization, DepthwiseConv2D
 from ..protocol.pond import PondPrivateTensor, PondMaskedTensor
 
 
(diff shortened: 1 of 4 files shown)


tf_encrypted/convert/register.py
```diff
@@ -177,6 +178,43 @@ def _keras_conv2d(converter, interiors, inputs):
...
+  except KeyError:
+    use_bias = False
+    bias_init = 'zeros'
+
+  shape = [i.size for i in kernel.attr["value"].tensor.tensor_shape.dim]
+
+  fmt = conv_op.attr["data_format"].s.decode('ascii')
+  fmt = "channels_last" if fmt == "NHWC" else "channels_first"
+
+  strides = int(max(conv_op.attr["strides"].list.i))
```
jvmancuso: curious, why do we take the max here?

avitalsh: @jvmancuso 
because when the stride is 2 then the conv op node looks like this - 
```
attr {
  key: "strides"
  value {
    list {
      i: 1
      i: 2
      i: 2
      i: 1
    }
  }
```

so using max we only extract the 2. 
This is how it was also done in the conv2d method in register.py

jvmancuso: hm, I think that should be fixed then (for conv2d as well).  we can do that in a separate PR though!

Issue #627, opened 2019-07-19
DepthwiseConv2D fixes


jvmancuso:
Recent PR #612 was merged but needs a few impovements -- the tests should be moved to convolutional_test.py and we should add a test for calling `set_weights` on the layer to ensure that works.

avitalsh:
Fixed in PR #631

PR #631, opened 2019-07-24
DWconv fix

avitalsh:
Following issue #627

Changes: 
1.  Changed ```DepthwiseConv2D``` class to be a child of the ```Conv2D``` class - this matches the implementation in tf.keras. This was needed because when calling ```model.get_config()``` the depthwise layer config dictionary includes the ```dilation_rate``` parameter and this raised an error when calling ```tfe.keras.models.model_from_config``` due to fact that the implemetation of depthwise not having a dilation_rate param.

2. Overloaded ```set_weights``` in the ```DepthwiseConv2D``` class - the weights in the depthwise layer work differently then conv layers. The kernel shape is ```[k, k, in_channels, 1]```. We rearrange the kernel to have the shape of ```[k, k, in_channels, in_channels]``` to be able to use it in the regular conv implementation. If we would use the set_weight method from the base layer we would have to do the rearrangement every time the layer is called. By overloading it, we can rearrange the kernel whenever it is changed. 

3. Moved the tests to convolutional_test.py.

**set_weight test** - wasn't sure what would be the best way to add such a test. I perfomed my own test in a logic similar to the tests for the sequential model - created a tf.keras model with depthwise conv layer, and created a tfe.keras model from the config and weights. This test works, but i'm not sure if this is the wanted logic here.


avitalsh:
@jvmancuso - I've added the test

PR #636, opened 2019-07-29
tf.keras global pooling support

avitalsh:
Added support for ```tf.keras.layers.GlobalAveragePooling2D``` and ```tf.keras.layers.GlobalMaxPooling2D``` layers, including converter support.

### DIFF
--- a/tf_encrypted/convert/register.py
+++ b/tf_encrypted/convert/register.py
@@ -10,7 +10,8 @@ import numpy as np
 import tensorflow as tf
 
 from ..layers import Conv2D, Relu, Sigmoid, Dense, AveragePooling2D, MaxPooling2D
-from ..keras.layers import BatchNormalization, DepthwiseConv2D, GlobalAveragePooling2D
+from ..keras.layers import BatchNormalization, DepthwiseConv2D, \
+  GlobalAveragePooling2D, GlobalMaxPooling2D
 from ..protocol.pond import PondPrivateTensor, PondMaskedTensor
 
 
@@ -55,6 +56,7 @@ def registry():
       "batch_normalization_v1": _keras_batchnorm,
       "depthwise_conv2d": _keras_depthwise_conv2d,
       "Mean": _global_avgpool,
+      "Max": _global_maxpool,
   }
 
   return reg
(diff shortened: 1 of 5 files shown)


tf_encrypted/convert/register.py
```diff
@@ -598,6 +601,37 @@ def _avgpool(converter, node: Any, inputs: List[str]) -> Any:
 
   return out
 
+def _global_avgpool(converter, node: Any, inputs: List[str]) -> Any:
```
jvmancuso: let's follow the trend of `_keras_{layer_identifier}` here

tf_encrypted/convert/register.py
```diff
@@ -598,6 +601,37 @@ def _avgpool(converter, node: Any, inputs: List[str]) -> Any:
...
+    data_format = 'channels_last'
+  else:
+    data_format = 'channels_first'
+
+  layer = GlobalAveragePooling2D(data_format=data_format)
+
+  return layer(x_in)
+
+
+def _global_maxpool(converter, node: Any, inputs: List[str]) -> Any:
```
jvmancuso: same as above

tf_encrypted/keras/layers/pooling.py
```diff
@@ -167,3 +167,104 @@ def __init__(self,
         self.prot.avgpool2d,
         pool_size=pool_size, strides=strides,
         padding=padding, data_format=data_format, **kwargs)
+
+class GlobalPooling2D(Layer):
```
jvmancuso: should this be a proper abstract base class from the `abc` module? in that case this would need to inherit from `abc.ABC` and `call` would need to be decorated with `@abstractmethod`

tf_encrypted/keras/layers/pooling.py
```diff
@@ -167,3 +167,104 @@ def __init__(self,
...
+          `(batch_size, channels, rows, cols)`
+
+  Output shape:
+      2D tensor with shape:
+      `(batch_size, channels)`
+  """
+
+  def call(self, inputs):
+    if self.data_format == 'channels_last':
+      x_reduced = self.prot.reduce_max(inputs, axis=2)
```
jvmancuso: is `reduce_max` not implemented as a tensor method? if so, we should open a separate issue to track that

avitalsh: it isn't implemented in pond, and when I called it using the tensor (inputs.reduce_max) i get an error that says that pond doesn't have reduce max

tf_encrypted/convert/register.py
avitalsh: @jvmancuso done

tf_encrypted/convert/register.py
avitalsh: @jvmancuso  done

tf_encrypted/keras/layers/pooling.py
avitalsh: @jvmancuso done

tf_encrypted/keras/layers/pooling.py
jvmancuso: I think this should be covered by our refactoring plans, so nvm on tracking with an issue

PR #637, opened 2019-07-29
Added conda installation instructions to README

VMS-6511:


README.md
```diff
@@ -13,6 +13,13 @@ TF Encrypted is available as a package on [PyPI](https://pypi.org/project/tf-enc
 ```bash
 pip3 install tf-encrypted
 ```
+Creating a conda environment to run TF Encrypted code can be done using:
```
ianlivingstone: What do we need to do to offer tf encrypted via conda? We should investigate how much work this is and the benefits, perhaps by opening a follow-on issue.

cc @mortendahl

PR #643, opened 2019-08-13
Fixed bugs in converter

avitalsh:
Following slack discussion with @jvmancuso I've fixed some bugs in the converter - 

1. In ```match_numbered_scope``` - added argument ```not_numbered``` to only match exact numbering. This was needed because when we had two layers of the same type, for example two batch norm layers, the second node - ```batch_normalization_v1_1``` - matched both nodes in the ```specop_dict``` in the convert method, and we only want to match the same node. 

2. In ```match_numbered_scope``` - changed the regex to match scoped nodes such as ```custom_layer/conv2d/kernel```.

3. In ```match_numbered_scope``` - changed the return mechanism to return the last group which is not none. This was needed because with the new regex, nodes such as ```conv2d_1/kernel``` returned the following groups - ```(None, None, 'conv2d_1', None)``` and this did not match the old return mechanism. 

4. In ```find_leaves``` - added an edge case for the ```batch_normalization_v1/keras_learning_phase``` node. This node serves as input to many other nodes but is irrelevant to the conversion to secure model and therefore was ignored to avoid errors.

### DIFF
--- a/tf_encrypted/convert/convert.py
+++ b/tf_encrypted/convert/convert.py
@@ -288,16 +288,19 @@ def find_leaves(scope, subscope_map):
 
 def match_numbered_scope(specop, search_string,
                          return_group=True,
-                         numbered=True
-                         ):
+                         numbered=True):
   """
   Find a numbered scope matching a specop from REGISTERED_SPECOPS,
   and return it if found.
     Example: 'conv2d' will match '...conv2d_345/...' and return 'conv2d_345'.
 
   Args:
+    specop: the specop to match.
+    search_string: the op name to search.
+    return group: if True, returns the last matching group, otherwise return
+        the match object.
     numbered: only match exact numbering  -
-            i.e. conv2d_1 will only match conv2d_1 and not conv2d etc.
+        i.e. conv2d_1 will only match conv2d_1 and not conv2d etc.
 
   """
   if numbered:
--- a/tf_encrypted/convert/convert.py
+++ b/tf_encrypted/convert/convert.py
@@ -94,7 +94,9 @@ class Converter:
         # Register high level special operations
         for s in specop_dict:
           # If this node is the output of the current specop, register it
-          if match_numbered_scope(s, node.name, return_group=False, not_numbered=True):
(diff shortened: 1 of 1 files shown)


tf_encrypted/convert/convert.py
```diff
@@ -279,21 +286,30 @@ def find_leaves(scope, subscope_map):
...
 
-def match_numbered_scope(specop, search_string, return_group=True):
+def match_numbered_scope(specop, search_string,
+                         return_group=True,
+                         not_numbered=False):
   """
   Find a numbered scope matching a specop from REGISTERED_SPECOPS,
   and return it if found
 
+  :param not_numbered: only match exact numbering  -
```
jvmancuso: Hey, we're actually moving toward [Google Style Docstrings](http://google.github.io/styleguide/pyguide.html#383-functions-and-methods)! Sorry for the confusion

tf_encrypted/convert/convert.py
```diff
@@ -279,21 +286,30 @@ def find_leaves(scope, subscope_map):
   return input_leaves, output_leaves
 
 
-def match_numbered_scope(specop, search_string, return_group=True):
+def match_numbered_scope(specop, search_string,
+                         return_group=True,
+                         not_numbered=False):
```
jvmancuso: To make this slightly more readable, I'd suggest renaming this to `numbered` and flipping the truth value to avoid the double negative here

tf_encrypted/convert/convert.py
avitalsh: @jvmancuso changed for the ```numbered``` param but wasn't sure what to write about the ```return_group``` param.

tf_encrypted/convert/convert.py
avitalsh: @jvmancuso Done.

tf_encrypted/convert/convert.py
jvmancuso: fixed, added some definitions in the docstring

PR #656, opened 2019-09-09
Fix edge case when using specific specops in Converter

avitalsh:
Following slack discussion with @jvmancuso:
Fixed bug in ```find_output_names``` - the problem was that it was searching for names in the trimmed_pb instead of the entire graph. This was not working for a model of the type - 
```
x1, x2 = ChannelSplit()(input_img) #tf.split wrapped as keras layer
x = tf.keras.layers.Conv2D(6, (2,2), use_bias=True, padding='same')(x1)
x = tf.keras.layers.Concatenate(axis=-1)([x2, x])
```
The first output of split was an input to the conv node which is removed in trimmed_pb, and therefore was not found.

This was fixed by searching the original graphdef instead of the trimmed one.

Another change in ```find_output_names``` was to insert the names by their order and not by the order of apperance in the graph. This caused bugs when the second output was used before the first, due to 
https://github.com/tf-encrypted/tf-encrypted/blob/92722ba13c9f536066e6269341ec8963041caef5/tf_encrypted/convert/convert.py#L132-L133

### DIFF
--- a/tf_encrypted/convert/convert.py
+++ b/tf_encrypted/convert/convert.py
@@ -82,8 +82,8 @@ class Converter:
     # Create a dictionary excluding all the sub ops related to required_space_to_batch_paddings
     # Except the sub ops related to the input or output of this special ops.
     pb_trimmed, graph_def = select_relevant_ops(specop_inputs,
-                                     specop_outputs,
-                                     graph_def)
+                                                specop_outputs,
+                                                graph_def)
     node_list = pb_trimmed.values()
 
     # If the ops are not related to the special ops, use the existing approach to register them.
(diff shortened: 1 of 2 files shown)


jvmancuso:
CI should be able to pass for this once we pull changes from master!

mortendahl:
this looks good to me; @jvmancuso any comments?

mortendahl:
any thoughts on what could be tested for @jvmancuso ?

jvmancuso:
I'm thinking a minimal example that reproduces the bug on master. A slimmed down version of the original code that caused this issue would work as well

avitalsh:
@jvmancuso test was added

Issue #658, opened 2019-09-12
OOM when running large exact ReLU layers 


avitalsh:
Hey, 
larger relu layers (for example 224x224x192) crash with OOM. 

example - 
```
input_img = tf.keras.layers.Input(shape=(224, 224, 192))
x = tf.keras.layers.ReLU()(input_img)
model = tf.keras.Model(inputs=input_img, outputs=x)

with tfe.protocol.SecureNN():
   s_model = secure_model(model)
   s_pred = s_model.private_predict(np.random.rand(1, 224, 224, 192))
```
Inspecting the model using tensorboard shows that the relu subgraph has 295 nodes, and the entire graph has 307. 

Following slack discussion with @mortendahl - i'll trying modifying the securenn relu layer so it will split and concat large tensors.

jvmancuso:
Closed by #661

PR #661, opened 2019-09-16
Fix for large relus OOM issue

avitalsh:
Following issue #658 - modified SecureNN's ReLU so that it will first check if size of tensor is above a predefined tensor. If so, the tensor is split to the minimal number of splits such that they are smaller then the threshold, ReLU is performed on each split and the results are then concatenated. 

@mortendahl - regarding the threshold: I started from the original tensor size that raised the problem in my model (224x224x192) and decreased channel size to find a size that won't crash. This is not necesarrily the "best" threshold. Do you have better idea to find threshold?

### DIFF
--- a/tf_encrypted/protocol/securenn/securenn.py
+++ b/tf_encrypted/protocol/securenn/securenn.py
@@ -340,7 +340,7 @@ class SecureNN(Pond):
     return self.dispatch('equal_zero', x, container=_thismodule, dtype=dtype)
 
   @memoize
-  def relu(self, x, max_size=224*224*95, **kwargs):
+  def relu(self, x, **kwargs):
     """
     relu(x) -> PondTensor
 
@@ -354,6 +354,7 @@ class SecureNN(Pond):
     :param PondTensor x: Input tensor.
     :param int max_size: max size of tensor that won't raise OOM
     """
+    max_size = kwargs.get('max_size', 224*224*95)
 
     def actual_relu(x):
       with tf.name_scope('relu'):
@@ -368,12 +369,11 @@ class SecureNN(Pond):
       max_channel = max_size // np.prod(shape[:-1])
       size_split = [max_channel] * int(shape[-1] / max_channel)
       if np.sum(size_split) < shape[-1]:
-          #there is a leftover from the orig tensor
-          size_split += [shape[-1] - np.sum(size_split)]
+        #there is a leftover from the orig tensor
+        size_split += [shape[-1] - np.sum(size_split)]
 
       x_split = self.split(x, size_split, axis=-1)
 
-      # for i in range(num_split):
       for i, _ in enumerate(x_split):
         x_split[i] = actual_relu(x_split[i])
 
--- a/tf_encrypted/protocol/securenn/securenn.py
+++ b/tf_encrypted/protocol/securenn/securenn.py
@@ -363,7 +363,7 @@ class SecureNN(Pond):
 
     total_size = np.prod(shape)
 
-    max_size = kwargs.get('max_size', 4000000)
+    max_size = kwargs.get('max_size', 2500000)
     if not max_size or total_size <= max_size:
       return actual_relu(x, 'relu')
 
--- a/tf_encrypted/protocol/securenn/securenn.py
+++ b/tf_encrypted/protocol/securenn/securenn.py
@@ -340,7 +340,7 @@ class SecureNN(Pond):
     return self.dispatch('equal_zero', x, container=_thismodule, dtype=dtype)
 
   @memoize
-  def relu(self, x):
+  def relu(self, x, max_size=224*224*95):
     """
     relu(x) -> PondTensor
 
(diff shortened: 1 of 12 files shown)


tf_encrypted/protocol/securenn/securenn.py
```diff
@@ -353,9 +353,38 @@ def relu(self, x):
...
+
+
+    max_size = 224*224*95
+    shape = x.shape.as_list()
+
+    if np.prod(shape) >= max_size:
+      #tensor is too big and will raise OOM
+
+      def check_num_split(num_split):
+        return (shape[-1] % num_split == 0 and
```
mortendahl: maybe add a quick comment about what informally satisfies the predicate

tf_encrypted/protocol/securenn/securenn.py
```diff
@@ -353,9 +353,38 @@ def relu(self, x):
...
-      drelu = self.non_negative(x)
-      return drelu * x
+
+    def relu(x):
+      with tf.name_scope('relu'):
+        drelu = self.non_negative(x)
+        return drelu * x
+
+
+    max_size = 224*224*95
```
mortendahl: this could be an optional parameter for `relu`, allowing experts to tweak as needed

tf_encrypted/protocol/securenn/securenn.py
```diff
@@ -353,9 +353,38 @@ def relu(self, x):
 
     :param PondTensor x: Input tensor.
     """
-    with tf.name_scope('relu'):
-      drelu = self.non_negative(x)
-      return drelu * x
+
+    def relu(x):
```
mortendahl: consider renaming (`actual_relu`, `sub_relu`, ?) to avoid any confusion

tf_encrypted/protocol/securenn/securenn.py
```diff
@@ -353,9 +353,38 @@ def relu(self, x):
...
+      #tensor is too big and will raise OOM
+
+      def check_num_split(num_split):
+        return (shape[-1] % num_split == 0 and
+                np.prod(shape[:-1] + [shape[-1] / num_split]) < max_size)
+
+
+      #look for best num_split value
+      num_split = 2
+      while not check_num_split(num_split):
```
mortendahl: looping like this might be a bit expensive, but since it all happens at compile time it might not be a big issue. however, i believe it could cause a livelock in the cases where it is not enough to only split the last dimension, ie if `shape[:-1] + [1]` is still too large.

an alternative approach would be to not insist on splitting into equal sized subtensors but instead take a number of maxsize tensors as possible and a single left-over tensor.

in any case, taking all dimensions into account is probably overkill for now, and think it's fine to only consider how to split the last dimension for now. it should raise an error instead of livelocking though.

avitalsh: I was considering flattening the tensor, splitting as you suggested, performing ReLU and then concatenating and reshaping to original shape. This way we don't limit the implementation to the last dimension, but this also adds the cost of the flattening and reshaping. @mortendahl What do you think?

tf_encrypted/protocol/securenn/securenn.py
avitalsh: done

tf_encrypted/protocol/securenn/securenn.py
avitalsh: done

tf_encrypted/protocol/securenn/securenn.py
avitalsh: done

tf_encrypted/protocol/securenn/securenn.py
avitalsh: @mortendahl this caused the CI the crash because now the secureNN relu params differ then pond's relu. This can be solved by adding this to pond as well, even though it will not be used there..

tf_encrypted/protocol/securenn/securenn.py
mortendahl: I would go for the simple solution for now and only split along the last dimension (passing a vector to `tf.split`), but raise an error in the cases where that's not enough.

If you prefer the fuller solution then I think flattening and reshaping is actually very cheap, as I believe they only change the view of the data, not how it's laid out in memory. Splitting and joining is the more expensive operations as they might copy data around.

tf_encrypted/protocol/securenn/securenn.py
mortendahl: what if we put it in the kwargs instead and add that to both? i'm thinking that this is a securenn specific implementation detail that we may not want to have as part of the public API.

tf_encrypted/protocol/securenn/securenn.py
jvmancuso: I'm not sure if this helps much, but `tf.split` accepts a 1d-array for the `axis` argument as well as the `num_size` argument, which allows it to split across multiple dimensions at once

PR #663, opened 2019-09-17
remove top-level calls to logging.basicConfig()

midokura-silvia:
Currently when calling `import tf_encrypted`, this implies a call to `logging.basicConfig()`. 
This function has the particularity, that only its first invocation is taken into account and any later calls are ignored. 

However, users of tf_encrypted (as for example users of PySyft) would like to be able to set the logging format themselves and not inherit the setting from tf_encrypted. 
Therefore, I propose to remove all top-level calls of `basicConfig()` and move them to the files acting as scripts (directly invoked, or having a section `if name == '__main__.py'`).

### DIFF
--- a/examples/bloom/__main__.py
+++ b/examples/bloom/__main__.py
@@ -18,9 +18,7 @@ genebanks = [
     DataOwner("genebank-6", NUM_FEATURES, TRAINING_SET_SIZE, TEST_SET_SIZE),
 ]
 
-logging.basicConfig()
-logger = logging.getLogger("tf_encrypted")
-logger.setLevel(logging.DEBUG)
+logging.basicConfig(level=logging.DEBUG)
 
 model = BloomRegressor()
 model.fit(genebanks)
--- a/examples/bloom/__main__.py
+++ b/examples/bloom/__main__.py
@@ -19,7 +19,7 @@ genebanks = [
 ]
 
 logging.basicConfig()
-logger = logging.getLogger(__name__)
+logger = logging.getLogger("tf_encrypted")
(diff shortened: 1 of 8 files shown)


mortendahl:
looks great! could we add a logging config to `tf_encrypted/player/__main__.py` as well?

examples/bloom/__main__.py
```diff
@@ -16,6 +18,10 @@
     DataOwner("genebank-6", NUM_FEATURES, TRAINING_SET_SIZE, TEST_SET_SIZE),
 ]
 
+logging.basicConfig()
+logger = logging.getLogger(__name__)
```
mortendahl: Using `__name__` here instead of `tf_encrypted` might prevent `logger.setLevel(logging.DEBUG)` below from having an effect. Same for the other files.

Alternatively it might be okay to simply do `logging.basicConfig(level=logging.DEBUG)` here, if it doesn't produce too much debugging information from other packages.

midokura-silvia: You're right, I usually set the level in the call to basicConfig. Probably best to do that. And if you have too much output, you can still increase the logging level from the other loggers. 
So which solution do you prefer?

mortendahl: would you mind giving `logging.basicConfig(level=logging.DEBUG)` a try and see if it prints a lot when run? if not then let's go with that.

mortendahl: gave it a try and it looks good; nothing extra printed

midokura-silvia:
> Happy to merge this one up if you're okay with the changes I pushed!

Perfect, go ahead!

PR #666, opened 2019-09-20
fix buggy implementation in interactive truncation

acs6610987:
The current implementation of interactive truncation is inconsistent with the paper, and might cause both information leakage and computation error.

As mentioned in the code, using `b = a + bound` will automatically lift `bound` by another scaling factor, hence making both `bound` and `b` TRIPLE-precision. This can cause two issues:

- `b` will be out of our intended range, and the security implication of this is unclear.
- The following revealed `c` will be larger than we expect, and might overflow the max int64 number (2^{63}-1), hence becoming negative, and then when we take the modulo `c %  scaling_factor`, it will be wrong. (The tests have not experienced this issue because the number is not big enough.)

### DIFF
--- a/tf_encrypted/protocol/pond/pond.py
+++ b/tf_encrypted/protocol/pond/pond.py
@@ -2382,8 +2382,11 @@ def _truncate_private_interactive(
     # assumption that the values originally lie in `[-B, B)`, and will
     # leak private information otherwise
 
+    # 'a + bound' will automatically lift 'bound' by another scaling factor,
+    # so we should first divide bound by the scaling factor if we want to
+    # use this convenient '+' operation.
     bound = prot.fixedpoint_config.bound_double_precision
-    b = a + bound
+    b = a + (bound / scaling_factor)
 
     # next step is for server0 to add a statistical mask to `b`, reveal
     # it to server1, and compute the lower part


mortendahl:
Nice catch! 💪

tf_encrypted/protocol/pond/pond.py
```diff
@@ -2382,8 +2382,11 @@ def _truncate_private_interactive(
     # assumption that the values originally lie in `[-B, B)`, and will
     # leak private information otherwise
 
+    # 'a + bound' will automatically lift 'bound' by another scaling factor,
+    # so we should first divide bound by the scaling factor if we want to
+    # use this convenient '+' operation.
     bound = prot.fixedpoint_config.bound_double_precision
-    b = a + bound
+    b = a + (bound / scaling_factor)
```
mortendahl: fixing it this way would leave the lower bits zero afaict due to the multiplication by the scaling factor. it might be better to avoid using the higher-level operations although and unwrap first thing.

mortendahl: it might be better to avoid messing with floating points although, and do everything directly on the lower-level tensors by unwrapping first thing.

acs6610987:
> lgtm!
> 
> > hence making both bound and b TRIPLE-precision
> 
> I don't see how `b` would end up being triple precision btw

Indeed, `a` is double-precision, `bound` is triple-precision, and I think the addition of the two would be triple-precision? Actually, maybe I should not use `precision` here, what I want to emphasize is the magnitude of the value. Both `b` and `bound` would be at the same magnitude as a triple-precision number?

mortendahl:
> I think the addition of the two would be triple-precision

we are unfortunately not tracking the precision level at this point (but would like to do so as part of a planned refactoring), only whether or not the value is scaled.

> Both `b` and `bound` would be at the same magnitude as a triple-precision number?

gotcha, yes, absolutely agree! thanks for clarifying.

PR #678, opened 2019-09-26
TensorFlow 2.0 first steps

justin1121:


tf_encrypted/protocol/pond/pond.py
```diff
@@ -221,7 +221,7 @@ def define_private_placeholder(
     factory = factory or self.tensor_factory
 
     suffix = "-" + name if name else ""
-    with tf.name_scope("private-placeholder{}".format(suffix)):
+    with tf.compat.v1.name_scope("private-placeholder{}".format(suffix)):
```
mortendahl: there seems to still be a [`tf.name_scope`](https://www.tensorflow.org/api_docs/python/tf/name_scope) in TF2 although it looks a bit different. did you play around with it to see what it would take to use it instead?

tf_encrypted/tensor/native.py
```diff
@@ -98,7 +98,7 @@ def sample_uniform(self,
       if secure_random.supports_secure_randomness():
         sampler = secure_random.random_uniform
       else:
-        sampler = tf.random_uniform
+        sampler = tf.random.uniform
```
mortendahl: when will the random values be sampled? when this line is executed or during session run?

tf_encrypted/protocol/pond/pond.py
justin1121: It works almost the same way but I think they made some minor changes so it wasn't safe to auto-convert it. Will look into it more to find out what these changes are.

tf_encrypted/tensor/native.py
justin1121: Pretty sure it's when the tf.function is actually called. Unless you are running in eager mode but I'm not sure we want to provide any security guarentees for eager mode, for now anyway.

mortendahl: > Unless you are running in eager mode

do we disable eager mode anywhere?

justin1121: Realized there aren't any uses of tf.function in this PR but I believe it should be the same for session runs.

Eager isn't explicitly disabled anywhere but it doesn't just work.

tf_encrypted/protocol/pond/pond.py
```diff
@@ -295,7 +298,72 @@ def define_public_variable(
         x_on_1 = factory.variable(v_on_1)
 
     x = PondPublicVariable(self, x_on_0, x_on_1, apply_scaling)
-    _initializers.append(x.initializer)
+    return x
+
+  def define_private_tensor(
```
mortendahl: I've been pondering this guy for a while. What was the motivation to introduce it here?

justin1121: Since `tf.function`s can be called multiple times while they're tracing and it only wants `tf.Variable`s to be initialized only once you need to have a check in your `tf.function` to make sure they're not initialized more than once. The recommended way is to the store the Variables inside of a class that created outside of the `tf.function`. Here's the example for the [docs](https://www.tensorflow.org/tutorials/customization/performance#variables):

```
class C: pass
obj = C(); obj.v = None

@tf.function
def g(x):
  if obj.v is None:
    obj.v = tf.Variable(1.0)
  return obj.v.assign_add(x)

print(g(1.0))  # 2.0
print(g(2.0))  # 4.0
```

So I just decided to make a `define_private_tensor` for now until we have a better solution. But also its not clear to my why we're using Variables all over the place when they're not really needed for a lot of our tests.

mortendahl: > So I just decided to make a `define_private_tensor` for now until we have a better solution. But also its not clear to my why we're using Variables all over the place when they're not really needed for a lot of our tests.

It makes sense to have it for now, but I think there's a larger discussion to be had before making this part of the public API.

The deeper issue is that TFE has mostly been designed around the idea that the script itself is simply setting up an encrypted computation between a set of parties, resulting in a graph that is safe for distribution by only describing how the computation is to be performed without containing any sensitive information. This fits with what we explicitly called the `master` earlier. The way of expressing the need to get data into the computation is via `tfe.local_computation` and `tfe.define_private_placeholder`.

However, in tests and [even some examples](https://github.com/tf-encrypted/tf-encrypted/tree/master/examples/notebooks/serving) it has proven useful or intuitive to sometimes treat the script as an active party in the computation, for instance as the client in an encrypted computation instead. In these cases the generated graph may contain sensitive values and is in general *not* safe for distribution. The current API tries to prevent this, resulting it the work around via variables that feels like a misuse.

On a technically level, the issue is that TFE is currently designed around in-graph computations with limited support for between-graph computations.

Moving forward we will need to find a way to accommodate both use cases, but I've previously avoided introducing a function like `define_private_tensor` since it may be confusing to the user that the value is *not* guaranteed to remain private. Two approaches for now might be:

1. update the examples (and perhaps even the tests) to actually do the right thing and provide inputs via `tfe.local_computation`

2. or, keep `define_private_tensor` but make a big fat note that this needs to be further addressed

justin1121: Agree with all that. Did not realize were using variables for a work around like this so feel like something needs to change with that API, guess that's what you're proposing with `tfe.local_computation`? Could we just get rid of `define_private_tensor` and `define_private_variable` (as public APIs anyway)?

mortendahl: > feel like something needs to change with that API, guess that's what you're proposing with `tfe.local_computation`?

We need to figure out how to be better at between-graph computations, as that'll likely be needed for TF Federated and PySyft-TF as well. I perhaps see this in addition in the existing `tfe.local_computation` though.

> Could we just get rid of `define_private_tensor` and `define_private_variable` (as public APIs anyway)?

We probably need to keep some notion of private variable around, although I think the API should be changed to something like `tfe.Variable`. `define_private_tensor` could go if there's an alternative way for you to quickly fix up the examples.

justin1121: > like `tfe.Variable.define_private_tensor` could go if there's an alternative way for you to quickly fix up the examples.

Lost me here. Typo?

mortendahl: I mean that we should probably keep `tfe.define_private_variable` but rename it to `tfe.Variable` to match better with TensorFlow (and generalize the notion of a variable, as it depends on the protocol and hence may not always be private).

If there's an alternative way for you to do what you're doing now without `define_private_tensor` then I'm up for removing it.

justin1121: Feel like the best way would be to use `tfe.local_computation` but I haven't looked into how that would work yet.

examples/federated-learning/func_lib.py
```diff
@@ -35,7 +35,7 @@ def reptile_model_fn(data_owner, iterations=3,
         zip(grads_k, data_owner.model.trainable_variables),
     )
 
-  return model.trainable_variables
+  return [var.read_value() for var in data_owner.model.trainable_variables]
```
mortendahl: why is the `read_value` needed? I vaguely remember that they changed the semantics in TF2 to not require this?

tf_encrypted/protocol/pond/pond.py
mortendahl: any updates here @justin1121?

examples/federated-learning/func_lib.py
justin1121: Its just because this function is wrapped in a local computation and is expecting a tensor to be returned. Without `read_value` a ResourceVariable is returned. We could fix it in the local computation implementation but this works to.

tf_encrypted/protocol/pond/pond.py
justin1121: Haven't really found good documentation explaining the differences. The auto-convert script says something about re-entering a scope via the name (v1 name scope implementation) but I don't think we're doing that anywhere anyway.

.circleci/config.yml
```diff
@@ -64,7 +64,7 @@ variables:
 jobs:
   build:
     docker:
-      - image: circleci/python:3.5-jessie
+      - image: circleci/python:3.5.7-stretch
```
mortendahl: I'm tempted to only support 3.6 or 3.7 in general -- wdyt?

mortendahl:
looks like a log file made it into the branch, probably shouldn't

.circleci/config.yml
justin1121: I think it'd be fine to just to support 3.7.

tf_encrypted/protocol/pond/pond.py
```diff
@@ -2136,9 +2193,12 @@ def __init__(self, prot, variable0, variable1, is_scaled):
     )
     self.variable0 = variable0
     self.variable1 = variable1
-    self.initializer = tf.group(
-        *[var.initializer for var in [variable0, variable1]]
-    )
+
+    # TensorFlow 2.0 variables don't have initializers??
```
mortendahl: what's the deal here? it looks like the initializer is still there, but automatically invoked in eager and graph mode

justin1121: In eager and with tf.functions the initializers aren't stored on the variables because they are `ResourceVariables`. With eager disabled and using a session they are stored here. Since we're still using sessions in some places we needed this check to take advantage of the initializers.

tf_encrypted/protocol/pond/pond.py
mortendahl: let's keep it for now and revisit later when we remove more of the compatibility functionality

PR #686, opened 2019-10-08
Debug Tensor equal

Liangrj5:
I find a bug in code, it should use tf.equal(value, -1) instead value == -1.


mortendahl:
Thank you @Liangrj5! Just for curiosity, did this produce an error message? Or simply incorrect behaviour?

Liangrj5:
It is just for curiosity, I don't get error when I run code in tfe.

Liangrj5:
It is just for curiosity, I don't get error when I run code in tfe.

PR #694, opened 2019-10-15
Tf2 cloud updates

justin1121:


yanndupis:
Hey @justin1121, 

Good work! I will let @mortendahl review this PR but just a quick observation. 

Part of this PR, you might want to update the README.md. in the gcp folder. For example do we still   need to create a `tfe-template`, do we need to run `gcloud compute ssh tfe-template < debian-install.sh` etc. probably no. 

Also might be good to give some pointer about how to login and create/set a gcp project in case people have limited experience with GCP. What do you think?

Thank you!

justin1121:
Thanks @yanndupis that's a good idea!

tools/gcp/create
```diff
@@ -1,6 +1,13 @@
...
-    $@ \
-    --image tfe-image \
-    --machine-type=n1-standard-4
\ No newline at end of file
+for instance in $@
+do
+    gcloud compute instances create-with-container $instance \
+        --container-image=docker.io/tfencrypted/tf-encrypted \
+        --container-command=python \
+        --container-arg="tf_encrypted/player/__main__.py" \
```
mortendahl: why not `-m tf_encrypted.player`?

tools/gcp/delete
```diff
@@ -1,5 +1,4 @@
 #!/bin/sh
 
-gcloud compute instances delete \
-    $@ \
-    --quiet
\ No newline at end of file
+gcloud compute instances delete $@ --quiet
```
mortendahl: we should loop here like in `create`

justin1121: You can pass multiple instances here and it will delete them all. Only reason we're not doing that for `create` is because we need to pass different args.

tools/gcp/create
justin1121: Might be another way to fix this but the command being run in the docker container wasn't being parsed correctly and causing the command to be chopped up into separate and not work. Think this will do for now.

PR #695, opened 2019-10-15
federated learning example updates

justin1121:


examples/federated-learning/run.py
```diff
@@ -8,44 +8,51 @@
...
+                    default=3,
+                    help="Specify how many data owners should "
+                         "take part in the learning. If remote "
+                         "configuration is specified the number of data owners "
+                         "must match the number specified here")
+parser.add_argument("--no-split", action="store_true", default=False,
+                    help="Whether the script runner should help "
+                         "simulate the training by splitting the data and "
+                         "distributing it amongst the data owners. "
+                         "Only applicable for local computations.")
```
jvmancuso: We should probably try to modify `split_dataset` to work with the remote config. I believe we can use TFRecordWriter in concert with [`tf.io.gfile.GFile`](https://www.tensorflow.org/api_docs/python/tf/io/gfile/GFile) to accomplish this

justin1121: Hmm does TFRecordWriter work with `tf.device` cause if so this should be pretty easy. E.g. Send tensor over gRPC and then write to file on that device. Was thinking that using `scp` or something like that to copy the tf record might be too complicated for this script but if we can just use `tf.device` it should be fairly straight-forward, either way we should probably do this in another PR.

PR #704, opened 2019-10-18
Fed learn secure

justin1121:


examples/federated-learning/players.py
```diff
@@ -139,6 +140,10 @@ def _update_one_round(self, data_owners, **kwargs):
 
     # Update step
     with self.device:
+      if isinstance(aggr_gradients[0], PondPrivateTensor):
```
mortendahl: I'd like to find a better way of addressing thus, eg not using reveal directly

justin1121: Sounds good. Suggestions welcome. Before we were using local computation for all of these functions but we decided to move away from that.

jvmancuso: this could also be grouped into `_call_aggregator_fn` to be applied automatically after the aggregation, but agree that it's currently too easy for the user to break security by not aggregating sufficiently in their custom `aggregator_fn` (assuming that's why @mortendahl isn't a fan, maybe I'm wrong)

mortendahl: > Before we were using local computation for all of these functions but we decided to move away from that.

what was the motivation behind this?

mortendahl: > it's currently too easy for the user to break security

security is already broken since the script running currently gets to see the gradients instead of just the model owner

mortendahl: I see it as a lower-lever primitive that makes it too easy to make mistakes: if you're not calling it while pinned to the right device you may end up leaking sensitive information to eg the script running (initially thought that happened here, until I saw the `pin_to_owner` specified several levels away).

justin1121: So what's the solution here? That the original was secure but not the ideal way of doing it? Should we just be using local_computation?

mortendahl: yeah I think it's secure (should be verified via the graph)!

justin1121:
Don't think this is needed anymore.

PR #708, opened 2019-10-23
fl without secure aggr

justin1121:


examples/federated-learning/func_lib.py
```diff
@@ -5,9 +5,11 @@
 
 from convert import decode
 
+from util import pin_to_owner
+
 ### Example model_fns ###
 
-def default_model_fn(data_owner):
+def _model_fn(data_owner):
```
jvmancuso: suggest calling this `_vanilla_model_fn` or `_keras_model_fn` or something to avoid clash with the base class's model_fn variable names

examples/federated-learning/players.py
```diff
@@ -97,9 +85,8 @@ def _call_aggregator_fn(self, grads, model):
   def _call_evaluator_fn(self):
     return self.evaluator_fn(self)
 
-  @tfe.local_computation
-  def call_model_fn(self, batches_or_owner):
-    return self.model_fn(batches_or_owner)
+  def call_model_fn(self, batches_or_owner, *args, **kwargs):
```
jvmancuso: suggest deleting these now, since the abstractmethods will be sufficient

jvmancuso: including the `_call_evaluator_fn` above & the BaseDataOwner's `call_model_fn`

examples/federated-learning/func_lib.py
justin1121: `_base_model_fn`?

examples/federated-learning/players.py
justin1121: I don't think removed `_call_evaluator_fn` makes sense we're still using pin to owner on that.

examples/federated-learning/main.py
```diff
@@ -85,18 +91,24 @@ class ModelOwner(BaseModelOwner):
                  representing the model owner.
   """
   @classmethod
```
mortendahl: why are these class methods btw?

examples/federated-learning/main.py
```diff
@@ -122,6 +134,11 @@ def build_data_pipeline(cls, local_data_file):
   # TODO: can move model_fn in here -- we leave it up to the user atm
 
 def main(_):
+  if not FLAGS.secure_aggr and FLAGS.reptile:
+    print("ERROR: Secure Aggregation must be used with "
```
mortendahl: why is that?

examples/federated-learning/main.py
```diff
@@ -122,6 +134,11 @@ def build_data_pipeline(cls, local_data_file):
...
+          "the reptile meta-learning algorithm.")
+    return
+
   if FLAGS.remote_config is not None:
     config = tfe.RemoteConfig.load(FLAGS.remote_config)
     config.connect_to_cluster()
```
mortendahl: we should probably change this its automatically called by `tfe.set_config`, say abstracted under `initialize` or something

examples/federated-learning/main.py
justin1121: @jvmancuso did this part!

examples/federated-learning/main.py
justin1121: https://github.com/tf-encrypted/tf-encrypted/issues/710

justin1121:
Yeah makes sense. I was thinking we'd just rewrite this once we have the abstractions figure out as a part of the FL RFC: https://github.com/tf-encrypted/rfcs/pull/1

examples/federated-learning/main.py
jvmancuso: I wrote them as classmethods in the base class, but they don't need to be classmethods in the subclass. in this case, the reason is that the model_fn is meant to be executed on the DataOwner and thus shouldn't depend critically on any state from the ModelOwner

examples/federated-learning/main.py
jvmancuso: would be pretty easy to do!

jvmancuso:
>I was thinking we'd just rewrite this once we have the abstractions figure out as a part of the FL RFC

same here!

examples/federated-learning/main.py
justin1121: done.

mortendahl:
let's get this merged and revisit as part of TF2 process. wdyt?

mortendahl:
@justin1121 and @jvmancuso any thoughts on what we should do with this one? solve the conflicts and merge?

mortendahl:
Closing, happy to reopen again as needed!

PR #709, opened 2019-10-24
Fl save model helper

justin1121:


examples/federated-learning/main.py
```diff
@@ -52,6 +52,9 @@
                       "simulate the training by splitting the data and "
                       "distributing it amongst the data owners. "
                       "Only applicable for local computations."))
+flags.DEFINE_string("save_dir", None,
+                    ("Where the model (h5 format) should be saved. If None the "
```
jvmancuso: suggest passing the format as an additional parameter that defaults to `h5` -- can change the name of this one to `save_path` to account for the difference between a saved model dir & h5 file in the different cases

examples/federated-learning/players.py
```diff
@@ -89,6 +90,9 @@ def _runner(self, data_owners, rounds, evaluate, evaluate_every, **kwargs):
         loss = self._call_evaluator_fn()
         prog.update(r, [("Loss", loss)])
 
+    # progress bar doesn't end with a newline XD
+    print()
```
jvmancuso: 🚀 🎸

examples/federated-learning/players.py
```diff
@@ -130,6 +134,18 @@ def _update_one_round(self, data_owners, **kwargs):
       self.optimizer.apply_gradients(zip(aggr_gradients,
                                          self.model.trainable_variables))
 
+  def save_model(self, save_dir):
+    if not os.path.exists(save_dir):
+      os.makedirs(save_dir)
```
jvmancuso: can change this block to `os.path.makedirs(save_dir, exist_ok)` to avoid a race condition between the calls to `exists` and `makedirs`

Issue #705, opened 2019-10-21
Error when converting a model with BatchNormalization layers


shreyansh26:
I trained a model with the following architecture.
```
model = Sequential()
model.add(Conv2D(32,(7,7), input_shape=(224, 224, 3), padding='same', activation='relu'))
model.add(MaxPooling2D(pool_size=(2, 2), strides=(2, 2)))
model.add(BatchNormalization())
model.add(Conv2D(64,(5,5), padding='same',activation='relu'))
model.add(MaxPooling2D(pool_size=(2, 2), strides=(2, 2)))
model.add(BatchNormalization())
model.add(Conv2D(128,(3,3), padding='same',activation='relu'))
model.add(MaxPooling2D(pool_size=(2, 2), strides=(2, 2)))
model.add(BatchNormalization())
model.add(Conv2D(128,(3,3), padding='same',activation='relu'))
model.add(MaxPooling2D(pool_size=(2, 2), strides=(2, 2)))
model.add(Flatten())
model.add(Dense(1000, activation='relu'))
model.add(Dense(2,activation='softmax'))
model.compile(optimizer='adam', loss = 'categorical_crossentropy', metrics=['accuracy'])
```
which works fine, after which I saved the model.
However during converting into a TFE model, first loading the model - 
```
batch_input_shape = (1, 224, 224, 3)
model = Sequential()
model.add(Conv2D(32,(7,7), batch_input_shape=input_shape, padding='same', activation='relu'))
model.add(MaxPooling2D(pool_size=(2, 2), strides=(2, 2)))
model.add(BatchNormalization())
model.add(Conv2D(64,(5,5), padding='same',activation='relu'))
model.add(MaxPooling2D(pool_size=(2, 2), strides=(2, 2)))
model.add(BatchNormalization())
model.add(Conv2D(128,(3,3), padding='same',activation='relu'))
model.add(MaxPooling2D(pool_size=(2, 2), strides=(2, 2)))
model.add(BatchNormalization())
model.add(Conv2D(128,(3,3), padding='same',activation='relu'))
model.add(MaxPooling2D(pool_size=(2, 2), strides=(2, 2)))
model.add(Flatten())
model.add(Dense(1000, activation='relu'))
model.add(Dense(2, name='logit'))

pre_trained_weights = 'model_batch_norm_1.hdf5'
model.load_weights(pre_trained_weights)

tf.reset_default_graph()
with tfe.protocol.SecureNN():
[...]

PR #711, opened 2019-10-26
Fix BatchNorm for model cloning

yanndupis:
Hey @mortendahl,

It's a WIP PR. This bug #705 is a bit more complicated than I expected. 

When we clone the a model, first we instantiate the model. During this process, the weights (TFE variables) from each layer gets stored in a `self.weights` model attribute. Then with `set_weights` we assign the numpy weights to the TFE variables.

The difference with Batchnorm is that the weights are not private variables but public variables (e.g moving_variance) because it involves sqrt, division etc. 

The problem is that we are less equipped to assign numpy weights (define_public_placeholder) to `PublicPondVariable`.

### DIFF
--- a/tf_encrypted/keras/layers/normalization.py
+++ b/tf_encrypted/keras/layers/normalization.py
@@ -184,6 +184,7 @@ class BatchNormalization(Layer):
             self.prot.add(moving_variance, self.epsilon)
         )
     )
+
     self.denom = denomtemp
 
     self.built = True
--- a/tf_encrypted/keras/layers/normalization.py
+++ b/tf_encrypted/keras/layers/normalization.py
@@ -1,8 +1,12 @@
 """Normalization layers implementation."""
+import numpy as np
+import tensorflow as tf
 from tensorflow.python.keras import initializers
 
 from tf_encrypted.keras.engine import Layer
 from tf_encrypted.keras.layers.layers_utils import default_args_check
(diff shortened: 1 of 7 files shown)


tf_encrypted/keras/layers/normalization.py
```diff
@@ -161,22 +165,26 @@ def build(self, input_shape):
...
+    self.moving_mean = self.add_weight(moving_mean, make_private=False)
 
     moving_variance = self.moving_variance_initializer(param_shape)
+    moving_variance = self.add_weight(moving_variance, make_private=False)
 
-    denomtemp = 1.0 / tf.sqrt(moving_variance + self.epsilon)
+    # to_navive() tranformation is bad and not yet working
+    # Find solution to compute tf.sqrt on PondPublicVariable
+    # or use different approach
+    denomtemp = 1.0 / tf.sqrt(moving_variance.to_native(), + self.epsilon)
```
yanndupis: really need to find a solution to this stuff. the challenge is that Keras weights is moving_variance when our public weight is `denomtemp` because of division/sqrt

yanndupis:
Hey @mortendahl, I think it's ready for review :). Thank you

yanndupis:
Thanks for the review! Sounds good, let's wait to settle on this.

yanndupis:
Most of the comments should be addressed. Thanks

PR #715, opened 2019-11-05
FL - Data Simulation

jvmancuso:
Allows users to run more TF Datasets with the FL example. Also cleans up the driving script a bit.

At one point I was trying to be able to do splitting/simulation of datasets in a generic way, but eventually conceded and limited this PR to image datasets (i.e. ones with `image` & `label` features).  This change occurred between commits `8a7b5ad` & `5242319`.


examples/federated-learning/players.py
```diff
@@ -28,20 +28,20 @@ class BaseModelOwner:
...
   """
 
-  def __init__(self, player_name, local_data_file, model, loss, optimizer=None):
+  def __init__(self, player_name, local_tfrecords, model, loss, optimizer=None):
     self.player_name = player_name
 
-    self.optimizer = optimizer
     self.loss = loss
+    self.optimizer = optimizer
+    self.dataset = tf.data.TFRecordDataset(local_tfrecords)
```
mortendahl: won't instantiation here not place the dataset on the wrong player?

examples/federated-learning/data_lib.py
```diff
@@ -0,0 +1,214 @@
...
+from functools import wraps
+import logging
+import os
+from pathlib import Path
+
+import tensorflow as tf
+import tensorflow_datasets as tfds
+import tf_encrypted as tfe
+
+logger = logging.getLogger('tf_encrypted')
```
mortendahl: maybe we should add this to core TFE, say `tfe.debugging.getLogger`?

jvmancuso:
>Okay to merge this up, but we need to define what we're trying to achieve with the FL example.

was deferring that to the TFF RFC, this work was meant to add capabilities for data scientist experimentation

examples/federated-learning/data_lib.py
jvmancuso: there is some precedent for doing so from TF, as they have their own debugger module. thinking we should still keep ours separate, but we should also try to standardize how we log things against theirs (currently there are some discrepancies). either way, think that calls for a separate PR :)

jvmancuso: there is some precedent for doing so from TF, as they have their own debugger module. thinking we should still keep ours separate, but we should also try to standardize how we log things against theirs (currently there are some discrepancies). either way, think that calls for a separate PR :)

mortendahl: See https://github.com/tf-encrypted/tf-encrypted/issues/722

PR #719, opened 2019-12-03
Bugfix: FL example broadcast leak

jvmancuso:
In the federated-learning example, there was a bug where data was leaking out of the TF graph during the broadcast step due to calls to Keras's `get_weights`/`set_weights`. This uses variable assignment to update the state on each data owner, pinning to each data_owner's device.

Warning: I wasn't able to get tensorboard graph viz working in TF 2.0, so I haven't been able to verify: (1) that this is the only leak, and (2) that this fixes the leak.

### DIFF
--- a/examples/federated-learning/players.py
+++ b/examples/federated-learning/players.py
@@ -80,11 +80,12 @@ class BaseModelOwner:
       self._update_one_round(data_owners, **kwargs)
 
       # Broadcast master model
+      # TODO: don't assume it's a tf.keras model
+      master_vars = self.model.trainable_variables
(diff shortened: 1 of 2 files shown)


mortendahl:
@jvmancuso think you may have the wrong base branch..? guessing it should be `master-tf2` instead?

jvmancuso:
oops

mortendahl:
You can change base branch if that's the case..

jvmancuso:
I don't think the failing build has to do with this PR, seems like a new dependency problem

PR #727, opened 2020-01-03
add loss BinaryCrossentropy_with_sigmoid and losses_test for BinaryCr…

qizhi-zhang:
I find there is a bug in  tf_encrypted/keras/losses.py: BinaryCrossentropy(Loss):
```
def grad(self, y_true, y_pred):
   return y_pred - y_true
```

Because 
`BinaryCrossentropy_loss(y_true, y_pred)= -(y_true log y_pred+(1-y_true)log (1-y_pred)), `
it implies the partial derivate of loss for y_pred should be  `-(y_true / y_pred - (1 - y_true) / (1 - y_pred))` but not `y_pred-y_true`. 

I think the reason of that the function `grad(self, y_true, y_pred)` in  class `BinaryCrossentropy` is defined as `y_pred - y_true` is that the division by type `PondPrivateVariable` is difficult. Hence I propose a new loss class
BinaryCrossentropy_with_sigmoid(Loss), which is defined as 
               ` -(y_true log sigmoid(y_pred)+(1-y_true)log (1- sigmoid(y_pred)))`
and its partial derivate for `y_pred` is `sigmoid(y_pred)-y_true`.

I give the  class `BinaryCrossentropy_with_sigmoid(Loss)` in tf-encrypted/tf_encrypted/keras/losses.py, and its unitTest in tf_encrypted/keras/losses_test.py


jvmancuso:
Thanks for the PR, and nice catch!  There is a story around this, despite it not yet being well documented.

In this loss function, we expect `y_pred` to already be a probability and not in the logit form, which means that the grad `y_pred - y_true` is actually equivalent to what you're suggesting.  In truth, the docstring [here](https://github.com/tf-encrypted/tf-encrypted/blame/fbbb23a59a3a374780cfc5d05d9cc53fea062035/tf_encrypted/keras/losses.py#L35) needs to be fixed to say `"""Computes the cross-entropy loss between true labels and predicted probabilities"""`.

So even though this isn't so much a bug as it is a documentation error, we _are_ missing a case in BinaryCrossentropy for when `from_logits=True` (and it looks like your new loss corresponds exactly to that case).  As such, it would be great if we can combine the two into one class, ideally following the API of the TF version [here](https://www.tensorflow.org/api_docs/python/tf/keras/losses/BinaryCrossentropy?version=stable).

I'd be happy to guide this PR through if you're open to making those changes!

qizhi-zhang:
> Thanks for the PR, and nice catch! There is a story around this, despite it not yet being well documented.
> 
> In this loss function, we expect `y_pred` to already be a probability and not in the logit form, which means that the grad `y_pred - y_true` is actually equivalent to what you're suggesting. In truth, the docstring [here](https://github.com/tf-encrypted/tf-encrypted/blame/fbbb23a59a3a374780cfc5d05d9cc53fea062035/tf_encrypted/keras/losses.py#L35) needs to be fixed to say `"""Computes the cross-entropy loss between true labels and predicted probabilities"""`.
> 
> So even though this isn't so much a bug as it is a documentation error, we _are_ missing a case in BinaryCrossentropy for when `from_logits=True` (and it looks like your new loss corresponds exactly to that case). As such, it would be great if we can combine the two into one class, ideally following the API of the TF version [here](https://www.tensorflow.org/api_docs/python/tf/keras/losses/BinaryCrossentropy?version=stable).
> 
> I'd be happy to guide this PR through if you're open to making those changes!
> 
> PS: In the future, you can run `make lint` and `make test` from the repo to make sure your code passes our style guide and that no tests are broken, which will make it much easier to get the PR through our CI.
 
Thanks for your comments!
Now, I have combine the  two class  `BinaryCrossentropy_with_sigmoid`and `BinaryCrossentropy` into the  class  `BinaryCrossentropy` using `from_logits` according to your comments. Now, the updated code is like 
```
if from_logits==True 
     loss(y_true, y_pred)= -(y_true log sigmoid(y_pred)+(1-y_true)log (1- sigmoid(y_pred))) 
     grad(y_true, y_pred)=sigmoid(y_pred)-y_true
else   (like original code)
    loss(y_true, y_pred)= -(y_true log y_pred+(1-y_true)log (1- y_pred))
    grad(y_true, y_pred)=y_pred-y_true
```


But I still have some concern about the code `grad(y_true, y_pred)=y_pred-y_true` in the case `from_logits==False`. In this case, 
`Loss= -(y_true log y_pred+(1-y_true)log (1- y_pred))`
and `y_pred=sigmoid(u)` for the output u of some Dense layer.
If we compute the  partial derivate of `loss` for `y_pred` as `y_pred-y_true`,  the partial derivate of `loss` for `u` will be computed as `(y_pred-y_true)*der_sigmoid(u)` in backward, it is not equal to the true partial derivate of `loss` for `u` :  `y_pred-y_true=sigmoid(u)-y_true`.

jvmancuso:
For this code, `grad` should always return the partial derivative of the loss with respect to `y_pred`.  Let's say `y_pred = p`, then we want the function to return `dL/dp`, and not `dL/du`, which is what you've written above.

For context, this is because we always expect the previous Keras layer to implement its own gradient function.  When `from_logits=False` here, then we expect the previous layer to be sigmoid, so that layer would be responsible for computing the sigmoid derivative to arrive at `dL/du` (which could then be used in the Dense layer before that to get the derivative for the weights).  If our autograd implementation were complete, this function would also calculate and return `dL/dy_true` as part of its code, but unfortunately we haven't implemented a generic autograd yet.

qizhi-zhang:
@jvmancuso   Thanks for your replay!
Please let me know if there is any need to change  in my updated code.
Morover, I find the following message in the Details of ci/circleci:

```
pylint tf_encrypted examples operations
************* Module tf_encrypted.protocol.ops_test
tf_encrypted/protocol/ops_test.py:594:8: E1121: Too many positional arguments for method call (too-many-function-args)
************* Module tf_encrypted.protocol.securenn.securenn_test
tf_encrypted/protocol/securenn/securenn_test.py:31:8: E1121: Too many positional arguments for method call (too-many-function-args)
tf_encrypted/protocol/securenn/securenn_test.py:42:8: E1121: Too many positional arguments for method call (too-many-function-args)
tf_encrypted/protocol/securenn/securenn_test.py:53:11: E1121: Too many positional arguments for method call (too-many-function-args)
tf_encrypted/protocol/securenn/securenn_test.py:121:10: E1121: Too many positional arguments for method call (too-many-function-args)
tf_encrypted/protocol/securenn/securenn_test.py:216:8: E1121: Too many positional arguments for method call (too-many-function-args)
......


```

but I did not modify these codes. Could you help me please?

qizhi-zhang:
> For this code, `grad` should always return the partial derivative of the loss with respect to `y_pred`. Let's say `y_pred = p` and `loss = L`, then we want the function to return `dL/dp`, and not `dL/du`, which is what you've written above.
> 
> For context, this is because we always expect the previous Keras layer to implement its own gradient function. When `from_logits=False` here, then we expect the previous layer to be sigmoid, so that layer would be responsible for computing the sigmoid derivative to arrive at `dL/du` (which could then be used in the Dense layer before that to get the derivative for the weights). If our autograd implementation were complete, this function would also calculate and return `dL/dy_true` as part of its code, but unfortunately we haven't implemented a generic autograd yet.


Hi @jvmancuso, 
I find there is the code of autograd in the lines 114~117 in https://github.com/tf-encrypted/tf-encrypted/blob/master/tf_encrypted/keras/models/sequential.py .     Why you say you haven't implemented a generic autograd ? 
And why we need to calculate `dL/dy_true` ?  I think that dy_true is not a variable. .

jvmancuso:
>Why you say you haven't implemented a generic autograd ?

We have autograd for Keras, but not for all of the lower-level ops in the `tf_encrypted` namespace.  So we can do autograd for all the Keras layers that have a backward pass implemented, but we can't do it generically for lower-level ops or a custom graph.  For example, there is no way to compute the derivative of `y = tfe.matmul(x1, x2)` without knowing that `tfe.matmul` was used and writing the derivative yourself in a function.

>And why we need to calculate `dL/dy_true` ? I think that dy_true is not a variable.

As far as I'm aware, the current best practice for autograd is to be able to run differentation for all nodes in a graph (whether they are plain tensors or variables). You can confirm this by checking that the `tf.gradients` function returns grads for Tensors, and not just Variables. This is because TensorFlow implements autodiff for its complete linear algebra library (and not just for the neural net helpers built on top).

jvmancuso:
I'm looking into the linting problem right now. It looks like `ndarray.reshape` expects a tuple of ints, but perhaps in a prior numpy version it also accepted an unrolled tuple of ints so pylint is only now breaking. Really not sure how that broke otherwise!

.idea/.gitignore
```diff
@@ -0,0 +1,3 @@
+
```
jvmancuso: Can you remove these additional files in the `.idea/` directory? I don't think they're necessary to make this change!

tf_encrypted/keras/losses_test.py
```diff
@@ -34,9 +36,49 @@ def test_binary_crossentropy(self):
...
+      loss = tf.keras.losses.BinaryCrossentropy(from_logits=True)
+      out = loss(y_true, y_pred)
+      der_for_y_pred = tf.sigmoid(y_pred)-y_true
+
       expected = sess.run(out)
+      expected_der = sess.run(der_for_y_pred)
 
     np.testing.assert_allclose(actual, expected, rtol=1e-1, atol=1e-1)
+    np.testing.assert_allclose(actual_der, expected_der, rtol=1e-1, atol=1e-1)
+    print("actual_der=", actual_der)
```
jvmancuso: I think the assert statements are sufficient, let's remove these prints if possible :)

tf_encrypted/keras/losses_test.py
```diff
@@ -34,9 +36,49 @@ def test_binary_crossentropy(self):
       y_pred = tf.convert_to_tensor(y_pred_np)
       loss = tf.keras.losses.BinaryCrossentropy()
       out = loss(y_true, y_pred)
+      der_for_y_pred = -(y_true / y_pred - (1 - y_true) / (1 - y_pred))
+
+      expected = sess.run(out)
+      expected_der = sess.run(der_for_y_pred)
+
+    np.testing.assert_allclose(actual, expected, rtol=1e-1, atol=1e-1)
+    print("actual_der=", actual_der)
```
jvmancuso: we can remove these two as well!

.idea/.gitignore
mortendahl: Addressed in #764

tf_encrypted/keras/losses_test.py
mortendahl: Addressed in #764

tf_encrypted/keras/losses_test.py
mortendahl: Addressed in #764

PR #735, opened 2020-01-16
ABY3 formatting

mortendahl:
This PR only reformats the ABY3 code. It uses `yapf` and the suggested config in https://github.com/tf-encrypted/tf-encrypted/issues/738.


tf_encrypted/tensor/boolfactory.py
```diff
@@ -8,323 +8,310 @@
...
+      self._maxval = maxval
+
+    @property
+    def shape(self):
+      return self._shape
+
+    @property
+    def value(self):
+      with tf.name_scope('expand-seed'):
+        return tf.cast(
```
mortendahl: not the best result here (with a `tf.bool` hiding at the end)

PR #739, opened 2020-01-16
Configuration for coding style

### LINKED ISSUE #738: Suggestions for improving code style tool support
We currently run `pylint` but there are several improvements that could be made, including better advice in `CONTRIBUTING.md` for how to use these.

Some suggestions:

- use `yapf` for code formatting with minor changes to the Google style
- use `flake8` (with extensions `flake8-mypy` and `flake8-docstrings`) and `pylint` for linting, (using both [seems somewhat common](https://www.reddit.com/r/Python/comments/82hgzm/any_advantages_of_flake8_over_pylint/)), adapting them to accept output of `yapf`
- use `setup.cfg` to configure everything, making it easy to align the various tools; this excludes pylint since it really wants its own file

See also https://github.com/tf-encrypted/tf-encrypted/issues/732 re advice on setting up a suitable development environment.

mortendahl:
Closes #738.

Missing:
- [ ] compare and potentially update linting rules against match current code base
- [ ] update development docs to reflect use of tools (not least `yapf`)
- [x] update makefile and CI to use configuration

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,19 @@
+[yapf]
+based_on_style = google
+indent_width = 2
+column_limit = 90
+split_before_dict_set_generator = True
+coalesce_brackets = True
+# split_before_first_argument = True
+
+[flake8]
+max-line-length = 90
+select = T,D,W,E,F
+ignore = E111,E114,E126,T484,T499,D202
+
+[pylint]
+max-default-line-length = 90
+
+[pydocstyle]
+
+[mypy]
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,18 +1,18 @@
 [yapf]
 based_on_style = google
 indent_width = 2
-column_limit = 90
+column_limit = 80
 split_before_dict_set_generator = True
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -0,0 +1,19 @@
+[yapf]
+based_on_style = google
+indent_width = 2
+column_limit = 90
```
jvmancuso: I believe this is a divergence from TensorFlow's `.pylintrc`, which we're currently using!  that has a conservative `column_limit = 79`.

if the change is intentional, it should be at least be better documented in the commit history and/or top-level for this PR. personally I like the character limit at 80, since it's _just right_ for viewing full scripts side-by-side on pretty much any modern screen

mortendahl: put it back to 80, let's go with that 👍

mortendahl:
@jvmancuso I suggest we merge this guy and then start a process of fixing the codebase and updating the rules until we're happy -- wdyt?

PR #743, opened 2020-01-24
Aby3 Improvement

zicofish:
Some important improvements to the ABY3 implementation:

1. Implement TFE's private while loop, so that for some algorithms, we can use `tfe.while_loop`, instead of Python's native for loop that creates unnecessary redundant graph nodes.
2. Update parallel prefix adder (`_B_ppa_private_private`)  to use `tfe.while_loop`. In my evaluation, this is at least twice faster than the previous version that uses Python's for loop.
3. Add indexer for ABY3PublicTensor for accessing the sub tensor of a ABY3PublicTensor
4. Some minor cleaning and renaming

### DIFF
--- a/tf_encrypted/protocol/aby3/aby3.py
+++ b/tf_encrypted/protocol/aby3/aby3.py
@@ -1052,9 +1052,9 @@ class ABY3(Protocol):
     return self.dispatch("B_not", x)
 
   @memoize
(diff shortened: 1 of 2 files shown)


zicofish:
This PR has some unresolved problem. I will take it back and submit another one later. Sorry :(

PR #753, opened 2020-02-12
Remove crypto_producer from Pond

Arash-Afshar:
Addresses https://github.com/tf-encrypted/tf-encrypted/issues/751

Removed `crypto_producer` and made `triple_source` accept values of multiple types instead of forcing all users of Pond to instantiate a `TripleSource`. Also, move the logic of gather_mask and negative_mask to `BaseTripleSource`. This way, Pond defers to triple_source for computing the operation using triple_source`s internal crypto producer.

btw, the test for `_indexer_masked` was not needed for this change. I was using it to make sure my tests actually work before writing tests for `_gather_masked` and `_negative_masked`. I kept it in since more tests are generally good ;)

Also, I wanted to use pytest `setUp` instead of my own `_setup` but I was not familiar with how tf handles graph contexts and if I can set it up once and use it later...

### DIFF
--- a/tf_encrypted/protocol/pond/pond_test.py
+++ b/tf_encrypted/protocol/pond/pond_test.py
@@ -5,9 +5,10 @@ import numpy as np
 import tensorflow as tf
 
 import tf_encrypted as tfe
-from tf_encrypted.protocol.pond import PondPublicTensor
+from tf_encrypted.protocol.pond import PondPublicTensor, PondMaskedTensor
 from tf_encrypted.tensor import int64factory, int100factory, native_factory
 from tf_encrypted.tensor import fixed100, fixed100_ni
+from .pond import _gather_masked, _indexer_masked, _negative_masked
 
 
 class TestPond(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


tf_encrypted/protocol/pond/triple_sources.py
```diff
@@ -104,6 +104,22 @@ def transpose_mask(self, a, perm):
 
     return a_t
 
+  def gather_mask(self, a, indices, axis):
```
jvmncs: For posterity, can you include a mention of this somewhere, either in the commit history or in the top-level comment of this PR?

Arash-Afshar: Updated the top level PR :)

tf_encrypted/protocol/pond/pond_test.py
```diff
@@ -156,6 +157,54 @@ def test_int100(self):
   def test_prime(self):
     self._core_test_sharing(native_factory(tf.int32, 67))
 
+class TestMasked(unittest.TestCase):
+
+  def _setup(self, dtype):
```
mortendahl: See https://github.com/tf-encrypted/tf-encrypted/issues/770

PR #763, opened 2020-02-13
update pytest marking

### LINKED ISSUE #762: Fix pytest marking
during `make test` we see  `PytestUnknownMarkWarning: Unknown pytest.mark.convert_maxpool - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/latest/mark.html`. seems like we may need to update syntax for this

jvmncs:
Fixes #762

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,8 @@
+[tool:pytest]
+markers =
+    slow: marks tests as slow (deselect with '-m "not slow"'),
+    convert_maxpool: marks the maxpool converter test, which is memory-intensive
+
 [yapf]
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -1,3 +1,8 @@
+[tool:pytest]
+markers =
+    slow: marks tests as slow (deselect with '-m "not slow"'),
+    convert_maxpool: marks the maxpool converter test, which is memory-intensive
```
mortendahl: longer than 80 chars :D

PR #765, opened 2020-02-13
VSCode setup

mortendahl:
Closes https://github.com/tf-encrypted/tf-encrypted/issues/732.


jvmncs:
wait this :green_salad: was a bit hasty -- how does this differ from the CONTRIBUTING.MD? maybe this should be a section in there?

mortendahl:
yeah, no, you're right. I thought this might be a longer doc used more as a reference, but if you're picking up a signal too then let's go with that!

PR #766, opened 2020-02-14
Tests, CI, linting, and formatting

mortendahl:
Main changes:

- fix unit tests and CI
- switch to building on Python 3.6
- docker images are no longer built
- pylint has been removed (leaving only flake8)
- all code is now formatting using black


tf_encrypted/convert/convert.py
```diff
@@ -351,7 +353,7 @@ def specop_from_numberedscope(scope):
 
 def strip_tensor_info(node_name: str) -> str:
   if node_name.startswith("^"):
-    return n[1:]
+    return node_name[1:]
```
mortendahl: @jvmncs does this look correct to you? was an on-the-fly bug fix

jvmncs: ya, I made this change as well!

tf_encrypted/convert/register.py
```diff
@@ -138,11 +146,13 @@ def _conv2d(converter, node, inputs):
...
-                 shape,
-                 strides=int(max(node.attr["strides"].list.i)),
-                 padding=node.attr["padding"].s.decode('ascii'),
-                 channels_first=fmt == "NCHW")
+  layer = Conv2D(
+      x_in.shape.as_list(),
+      shape,
+      strides=int(max(node.attr["strides"].list.i)),
+      padding=node.attr["padding"].s.decode('ascii'),
+      channels_first=fmt == "NCHW"
```
jvmncs: any reason why yapf isn't formatting these last lines in a standard way? sometimes it adds commas (which I think is preferable), but some like this one are missing one

tf_encrypted/convert/register.py
```diff
@@ -308,13 +323,17 @@ def _strided_slice(converter, node: Any, inputs: List[str]) -> Any:
...
   strides = tf.constant(strides.attr["value"].tensor)
 
-  return tfe.strided_slice(input_out, begin, end,
-                                          strides=strides,
-                                          begin_mask=begin_mask,
-                                          end_mask=end_mask,
-                                          ellipsis_mask=ellipsis_mask,
-                                          new_axis_mask=new_axis_mask,
-                                          shrink_axis_mask=shrink_axis_mask)
+  return tfe.strided_slice(
```
jvmncs: wow this is a huge improvement, not sure how this one was so bad before

tf_encrypted/keras/layers/convolutional.py
```diff
@@ -438,16 +475,24 @@ def compute_output_shape(self, input_shape):
...
+    )
     for d in range(self.depth_multiplier):
       for i in range(in_channels):
         mask[:, :, i, i + (d * in_channels)] = 1.
     return np.transpose(mask, [0, 1, 3, 2])
 
   def set_weights(self, weights, sess=None):
-    """ Sets the weights of the layer.
+    """
+    Sets the weights of the layer.
```
jvmncs: Is this recommended? I seem to recall Google docstrings having the first (imperative) sentence inline with the opening of the quotation

tf_encrypted/keras/layers/convolutional.py
```diff
@@ -438,16 +475,24 @@ def compute_output_shape(self, input_shape):
     return [n_x, n_filters, h_out, w_out]
 
   def get_mask(self, in_channels):
-    mask = np.zeros((self.kernel_size[0],
-                     self.kernel_size[1],
-                     in_channels, in_channels * self.depth_multiplier))
+    """TODO"""
```
jvmncs: if these new TODOs are common we should spin a "good first issue" out for them

tf_encrypted/keras/layers/normalization.py
```diff
@@ -84,62 +85,83 @@ class BatchNormalization(Layer):
...
-    default_args_check(renorm_clipping,
-                       "renorm_clipping",
-                       "BatchNormalization")
-    default_args_check(virtual_batch_size,
-                       "virtual_batch_size",
-                       "BatchNormalization")
-    default_args_check(adjustment,
-                       "adjustment",
-                       "BatchNormalization")
+        moving_variance_initializer
```
jvmncs: another example of a missing comma

jvmncs: seems strange that it happens so sporadically

tf_encrypted/keras/engine/base_layer.py
```diff
@@ -114,7 +116,7 @@ def add_weight(self, variable, make_private=True):
     return variable
 
   def set_weights(self, weights, sess=None):
-    """ Sets the weights of the layer.
+    """Sets the weights of the layer.
```
jvmncs: According to Google style guide, this should be imperative "Set" and there should be a newline at the end of this line, I think?

setup.cfg
```diff
@@ -6,16 +6,24 @@ markers =
...
-split_before_dict_set_generator=True
-coalesce_brackets=True
-# split_before_first_argument=True
+dedent_closing_brackets=True
+split_all_top_level_comma_separated_values=True
+split_before_first_argument=True
+split_penalty_after_opening_bracket=0
 
 [flake8]
 max-line-length=80
```
jvmncs: I thought you were going to increase this? I'm cool with it like this, but just curious :)

tf_encrypted/keras/models/__init__.py
```diff
@@ -5,9 +5,8 @@
 from .sequential import model_from_config
 from .sequential import clone_model
 
-
 __all__ = [
     'Sequential',
     'model_from_config',
-    'clone_model'
+    'clone_model',
```
jvmncs: aaaaaaand here's an example where it was added??

tf_encrypted/convert/register.py
mortendahl: yapf doesn't add commas on its own, you have to manually do that, which in turn forces this formatting. however, it may choose to go with this formatting either way if the line will otherwise become too long. the Google style is less likely to do this but the Facebook style does it by default; we are somewhere in-between thanks in part to `dedent_closing_brackets=True`).

with black on the other hand, there doesn't seem to be a way to force it to use this formatting, and it may decide to keep everything on a single line after one line break, which is not ideal either imho.

tf_encrypted/convert/register.py
mortendahl: yeah, column alignment pure sucks :D

tf_encrypted/keras/layers/convolutional.py
mortendahl: yeah, docstrings need a do-over at some point; luckily flake8 seems to capture them (but currently ignored in `setup.cfg`)

tf_encrypted/keras/layers/normalization.py
mortendahl: with yapf it has to be done by hand

tf_encrypted/keras/engine/base_layer.py
mortendahl: Believe you're right, let's fix in another issue

setup.cfg
mortendahl: think we should stick to standard as much as possible, and since we'll be using a formatter I think it's less of an issue

tf_encrypted/convert/register.py
jvmncs: yeah, that seems strange given black's stated goal of reducing diff churn. overall I'm okay with both if you are!

mortendahl:
@jvmncs notice that I've now removed pylint! perhaps there's a way to salvage it, but at least the current cont didn't play nicely with black.

Makefile
```diff
@@ -100,10 +89,19 @@ SPHINX_BUILD_GOOGLE_DOCSTRINGS = sphinx-apidoc
...
+
+BUILD_CONVERTER_README=$(CONVERT_DIR)/gen/readme_template.md
+$(BUILD_CONVERTER_README): $(BUILD_RESERVED_SCOPES) pythoncheck
+	python -m tf_encrypted.convert.gen.generate_reserved_scopes
+
 google-docstrings:
 	@$(SPHINX_BUILD_GOOGLE_DOCSTRINGS) -fMeET "$(SPHINX_PROJECT_DIR)" -o "$(SPHINX_NAPOLEAN_BUILD_DIR)"
 
-docs: google-docstrings
+docs: $(BUILD_CONVERTER_README) google-docstrings
```
jvmncs: I know it's technically correct, but I don't really like the converter readme being built with the defunct Sphinx docs. this makes it really easy for people to forget to build it

examples/house-credit-default/main.py
```diff
@@ -38,140 +46,153 @@
...
-  """Build a logistic regression model with tf.keras."""
-  model = keras.Sequential([
-      layers.Dense(1, use_bias=False, activation='sigmoid',
-                   input_shape=[input_shape]),
-  ])
+    """Build a logistic regression model with tf.keras."""
+    model = keras.Sequential(
+        [
+            layers.Dense(
+                1, use_bias=False, activation="sigmoid", input_shape=[input_shape]
```
jvmncs: it's times like these that black is a little excessive :D

examples/house-credit-default/main.py
```diff
@@ -38,140 +46,153 @@
...
-  train_y = train_y_df[1:246005]
+    # first 80 percent for training
+    train_x = train_x_df[1:246005]
+    train_y = train_y_df[1:246005]
 
-  # other 20 percent for evaluating
-  eval_x = train_x_df[246006:len(train_x_df) - 1]
-  eval_y = train_y_df[246006:len(train_y_df) - 1]
+    # other 20 percent for evaluating
+    eval_x = train_x_df[246006 : len(train_x_df) - 1]
```
jvmncs: spacing around a colon seems strange :/

examples/simple-average/run.py
```diff
@@ -4,47 +4,60 @@
...
+        ),  # pylint: disable=unexpected-keyword-arg
+        provide_input(
+            player_name="inputter-2"
+        ),  # pylint: disable=unexpected-keyword-arg
+        provide_input(
+            player_name="inputter-3"
+        ),  # pylint: disable=unexpected-keyword-arg
+        provide_input(
+            player_name="inputter-4"
+        ),  # pylint: disable=unexpected-keyword-arg
```
jvmncs: we can remove these linter statements now, right?

examples/house-credit-default/main.py
mortendahl: indeed :)

examples/house-credit-default/main.py
mortendahl: i vaguely recall them having a discussion re this somewhere if you're interested

examples/simple-average/run.py
mortendahl: yes!

Makefile
mortendahl: what if we let `build` depend on `docs`?

jvmncs: that sounds good to me!

mortendahl: Addressed in https://github.com/tf-encrypted/tf-encrypted/issues/779

PR #782, opened 2020-03-27
libsodium primitives for secure channels

mortendahl:
This adds a new distribution, `tf-encrypted-primitives` that contain ` sodium` module through which (select functionality of) [libsodium](https://libsodium.gitbook.io/doc/) can be accessed. For now this is not linked with main `tf-encrypted`, primarily because they target respectively TensorFlow 2 and 1.

See build instructions in `/primitives/README.md`.


tf_encrypted/primitives/sodium/easy_box_test.py
```diff
@@ -0,0 +1,62 @@
...
+class TestEasyBox(unittest.TestCase):
+    def setUp(self):
+        tf.reset_default_graph()
+
+    def test_gen_keypair(self):
+        pk, sk = easy_box.gen_keypair()
+        assert isinstance(pk, easy_box.PublicKey), type(pk)
+        assert isinstance(sk, easy_box.SecretKey), type(sk)
+
+        with tf.Session() as sess:
```
yanndupis: We will have to validate how easy to integrate this in TFF with TF 1.x version. Wonder if instead we should update this custom ops implementation to TF2.x and have it in separate branch for now. Might be easier to debug TFF with TF2.x.

cc: @jvmncs

primitives/README.md
```diff
@@ -0,0 +1,30 @@
+## Installation
+
+These commands must be run *in the root directory*, i.e. the parent of this one.
+
+To set up TensorFlow (one time requirement):
+
+```
+./configure
```
yanndupis, line 8: it should be renamed to `./configure.sh`.

tf_encrypted/primitives/sodium/easy_box_test.py
mortendahl: good point @yanndupis 👍 went with separate distribution for now, `tf-encrypted-primitives`, that only works with TF2 (could easily be made to work for TF1 as well)

primitives/README.md
mortendahl: good catch!

PR #788, opened 2020-04-06
shape inference functions for select ops

mortendahl:


yanndupis:
Hey @mortendahl,

Made progress however there is one big thing missing. I haven't found yet how to get the plaintext type to infer the shape in `op.cc`. Maybe I should create another context like OpKernel?

thanks,

primitives/tf_encrypted/primitives/sodium/cc/ops.cc
```diff
@@ -23,7 +41,15 @@ REGISTER_OP("SodiumEasyBoxSealDetached")
     .Input("sk_sender: uint8")
     .Output("ciphertext: uint8")
     .Output("mac: uint8")
-    .SetIsStateful();
+    .SetIsStateful()
+    .SetShapeFn([](shape_inference::InferenceContext* c){
+        shape_inference::ShapeHandle plaintext_shape = c->input(0);
+        shape_inference::ShapeHandle dtype_shape = c->MakeShape({sizeof(float)});
```
yanndupis, line 47: `sizeof(float)}` is hard coded, it should be inferred based on plaintext dtype.

mortendahl: pushed a fix

primitives/tf_encrypted/primitives/sodium/python/easy_box_test.py
```diff
@@ -55,20 +82,15 @@ def test_open_uint8(self):
         assert plaintext_recovered.shape == plaintext.shape
         np.testing.assert_equal(plaintext_recovered, np.array([[1, 2], [3, 4]]))
 
-    def test_open_float(self):
-        pk_s, sk_s = easy_box.gen_keypair()
-        pk_r, sk_r = easy_box.gen_keypair()
 
-        plaintext = tf.constant([1, 2, 3, 4], shape=(2, 2), dtype=tf.float32)
+def run_op(op, mode):
```
mortendahl: I've generalized this a tiny bit, see my commit

primitives/tf_encrypted/primitives/sodium/python/easy_box_test.py
```diff
@@ -1,50 +1,77 @@
...
+            ciphertext, mac = easy_box.seal_detached(plaintext, nonce, pk_r, sk_s)
+            plaintext_recovered = easy_box.open_detached(
+                ciphertext, mac, nonce, pk_s, sk_r, plaintext.dtype
+            )
+
+        assert ciphertext.raw.shape == plaintext.shape + (4,)
+        assert plaintext_recovered.shape == plaintext.shape
 
-    def test_seal_floats(self):
+    @parameterized.named_parameters(
```
mortendahl: nice, great call adding this!

mortendahl:
> Made progress however there is one big thing missing. I haven't found yet how to get the plaintext type to infer the shape in `op.cc`. Maybe I should create another context like OpKernel?

@yanndupis pushed a fix for this

primitives/tf_encrypted/primitives/sodium/python/easy_box_test.py
yanndupis: Nice! I didn't know `contextlib.suppress()`. Looks great!

yanndupis:
Ah wait, I forgot to set the shape for Mac.

yanndupis:
Hey @mortendahl, 

I fixed the shape for Mac. 

However I got a little surprise on the federated side. Even though the [encryption computation](https://github.com/tf-encrypted/federated/blob/6d752d94e819b72e033925a31d922070613caffd/tensorflow_federated/python/core/impl/executors/federating_executor.py#L636) returned tensors with defined shape, the [decryption computation](https://github.com/tf-encrypted/federated/blob/6d752d94e819b72e033925a31d922070613caffd/tensorflow_federated/python/core/impl/executors/federating_executor.py#L656) when defined has input tensors with empty shape (shape=(), not shape=unkown). Because ciphertext has a rank = 0, I was getting an error [here](https://github.com/tf-encrypted/tf-encrypted/blob/6e1bbfeb900a152beb94791df6f38876daed6172/primitives/tf_encrypted/primitives/sodium/cc/ops.cc#L78) when setting the recovered_plaintext shape. So I added a control flow to set the `plaintext_recovered shape = ()` when `ciphertext.shape.rank = 0 `. This happens when we define the computation but pretty sure that the shapes get updated at runtime (when the graph gets executed).

Thank you

primitives/tf_encrypted/primitives/sodium/cc/ops.cc
```diff
@@ -33,4 +67,15 @@ REGISTER_OP("SodiumEasyBoxOpenDetached")
...
     .Input("sk_receiver: uint8")
     .Output("plaintext: plaintext_dtype")
-    .SetIsStateful();
+    .SetIsStateful()
+    .SetShapeFn([](shape_inference::InferenceContext* c){
+        shape_inference::ShapeHandle ciphertext_shape = c->input(0);
+        shape_inference::ShapeHandle plaintext_shape;
+        auto rank = c->Rank(ciphertext_shape);
+        if (rank == 0)
+            plaintext_shape = c->MakeShape({});
```
mortendahl, line 76: perhaps it's better not to set a shape in this case, ie you could rewrite this as follows:

```
if (rank >= 1) {
  TF_RETURN_IF_ERROR(c->Subshape(ciphertext_shape, 0, rank-1, &plaintext_shape));
  c->set_output(0, plaintext_shape);
}
```

primitives/tf_encrypted/primitives/sodium/cc/ops.cc
```diff
@@ -23,7 +42,22 @@ REGISTER_OP("SodiumEasyBoxSealDetached")
     .Input("sk_sender: uint8")
     .Output("ciphertext: uint8")
     .Output("mac: uint8")
-    .SetIsStateful();
+    .SetIsStateful()
+    .SetShapeFn([](shape_inference::InferenceContext* c){
+        shape_inference::ShapeHandle mac_shape = c->MakeShape({crypto_box_MACBYTES});
+        c->set_output(1, mac_shape);
```
mortendahl, line 48: I would keep these in order, ie move logic for mac shape below logic for ciphertext shape

mortendahl:
btw, is there any reason you're not simply using `@tff.tf_computation` instead of `@computations.tf_computation`? the latter might make the reader wonder if there's any difference.

primitives/tf_encrypted/primitives/sodium/cc/ops.cc
mortendahl: no forget what I said; your solution is better 👍

primitives/tf_encrypted/primitives/sodium/cc/ops.cc
mortendahl: sorry, pushing this change :D

primitives/tf_encrypted/primitives/sodium/cc/ops.cc
mortendahl: Okay maybe the story doesn't end here after all.

AFAICT tensors of shape `()` are scalars, and ciphertext scalars only make sense for scalar `uint8` plaintexts. [This line](https://github.com/tf-encrypted/federated/blob/master/tensorflow_federated/python/core/api/computation_types.py#L101) in TFF seems a bit fishy to me since `tf.TensorShape([]) != tf.TensorShape(None)`, but I could be missing an equivalence here (cc @jvmncs any thoughts?).

At any rate, perhaps it's worth being a bit more explicit about the type signatures for our `tf_computations`, using e.g. `tf.TensorSpec(tf.uint8, None)` instead of simply `tf.uint8`. Let me test if this solves the problem you ran into.

Moreover, perhaps it's better to use `c->RankKnown` instead of testing against `c->Rank()`. In any case, if the rank is unknown then `c->Rank()` returns -1, which will caused problems if we're checking `== 0` instead of flipping branches and using `>= 1`.

mortendahl:
one more thing: consider using `tff.tf2_computation` instead of `tff.tf_computation`

primitives/tf_encrypted/primitives/sodium/cc/ops.cc
jvmncs: > but I could be missing an equivalence here

They're definitely not equivalent in general. tf.TensorShape([]) denotes scalar, whereas tf.TensorShape(None) is essentially a shapeless placeholder (e.g. user-defined). It seems to me like this is a restriction TFF introduces that deviates from the TF definition. Note they still allow for passing a `TensorShape(None)` object to that constructor.

mortendahl: I think I'm close to being at the bottom of this. Using e.g. `tf.uint8` as a type argument is indicating a scalar, and it should indeed be something like `tf.TensorSpec(dtype=tf.uint8, shape=None)` instead.

jvmncs: Yep, in that case I think you'd want something like `tff.TensorType(tf.uint8, tf.TensorShape(None))`

mortendahl: yes, or simply `tff.TensorType(tf.uint8, None)` (note that `tff.to_type(tf.TensorSpec(dtype=dtype, shape=shape)) == tff.TensorType(dtype, shape)`

PR #796, opened 2020-04-17
fix formatting

mortendahl:


primitives/tf_encrypted/primitives/sodium/python/easy_box_test.py
```diff
@@ -1,12 +1,12 @@
 # pylint: disable=missing-docstring
 import unittest
-from absl.testing import parameterized
 
 import numpy as np
 import tensorflow as tf
+from absl.testing import parameterized
```
jvmncs, line 6: do we have an issue about adding `isort`? that'd be good to have

mortendahl: what do you mean by "adding"?

just realized that adding `flake8-isort` works, so did that.

jvmncs: meant adding it to CI, which it looks like you've done in Makefile here

mortendahl: 👍

PR #816, opened 2020-07-22
support insecure randomness in ABY3 for research uses

zicofish:
After collecting experience from some users, the existing strict requirement of "secure randomness" in ABY3 is a little annoying because the dependency `libsodium` library is quite slow to download in some areas.

In this commit, I have added the support for "insecure randomness" in ABY3, just like that in Pond and SecureNN. It no longer forces users to install `libsodium` before using ABY3.


zicofish:
linting failed. Let me first fix the format stuff...

PR #820, opened 2020-08-04
Update TFE Primitives to new TF Big API

mortendahl:
See https://github.com/tf-encrypted/tf-big/pull/74.

Original motivation for this was to avoid dealing with bitlength in TFE Primitives.

### DIFF
--- a/primitives/tf_encrypted/primitives/paillier/primitives.py
+++ b/primitives/tf_encrypted/primitives/paillier/primitives.py
@@ -68,7 +68,7 @@ class Randomness:
     def __init__(self, raw_randomness: tf.Tensor):
         self.raw = _import_maybe_limbs(raw_randomness)
 
-    def export(self, dtype: tf.DType = tf.string) -> tf.Tensor:
+    def export(self, dtype: tf.DType = tf.uint8) -> tf.Tensor:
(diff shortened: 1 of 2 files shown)


primitives/tf_encrypted/primitives/paillier/primitives.py
```diff
@@ -35,49 +50,52 @@ def __init__(self, p, q):
...
 class Ciphertext:
-    def __init__(self, ek: EncryptionKey, raw_ciphertext):
+    def __init__(self, ek: EncryptionKey, raw_ciphertext: tf.Tensor):
         self.ek = ek
-        self.raw = tf_big.convert_to_tensor(raw_ciphertext)
+        self.raw = _import_maybe_limbs(raw_ciphertext)
 
-    def export(self, dtype: tf.DType = tf.string):
-        return tf_big.convert_from_tensor(self.raw, dtype=dtype)
+    def export(self, dtype: tf.DType = tf.string) -> tf.Tensor:
```
jvmncs, line 84: any reason why `tf.string` is used here and `tf.uint8` is used above?

mortendahl: nice catch! no, we should probably use tf.unit8 everywhere, will push a fix

PR #868, opened 2022-11-10
TFE based on TF2

zjn-code:
tfe based on tf2
- eager execution
- auto add dependency for stateful op
- define a graph by `tfe.function`
- nearly the same performance as tf1
- build tfe model from onnx model
- fix device bug


zjn-code:
aby3 test error cost too much time

PR #883, opened 2023-05-09
Suggest install from source to avoid problems

vincehong:


zjn-code:
TF encrypted needs compiled binary library to support custom op. But it seems like tensorflow doesn't have ABI compatibility between different minor version, e.g. tf2.9.x and tf2.10.x doesn't have ABI compatibility. I'm planning to support only one minor version(2.9.x) for pip package, and suggest users who use other versions of tensorflow to install from source.

Issue #900, opened 2024-09-23
Security Vulnerability in tf-encrypted Secure Random Module


GuopengLin:
## Description:
I have identified a critical security vulnerability in the random number generation module in tf-encrypted.

When the types of seed, minval, and maxval differ, the module generates a constant value instead of random numbers.

## Example:
In the following example, the result (res) is always a constant tensor of [-128, -128] (even the seed is changed), instead of a uniformly random tensor within the specified range.

```
import tensorflow as tf
from tf_encrypted.operations.secure_random.secure_random import secure_random_module

seed = tf.constant([-27587587, -1632666452, -218620675, -1537614956, 1554788191, -1805869028, -74859994, -1320420251], dtype=tf.int32)
shape = (2,)
minval = tf.constant(-128, dtype=tf.int8)
maxval = tf.constant(127, dtype=tf.int8)

res = secure_random_module.secure_seeded_random_uniform(
    shape,
    seed,
    minval,
    maxval,
    name=None,
)
print("res = ", res)  # Output: [-128, -128]

```

## Impact:
This vulnerability directly affects secure comparison protocols in aby3,  making them insecure. Below is a call stack showing how this vulnerability can be triggered:

## Call Stack:

```

File "~/tf-encrypted/./test_rss_eq.py", line 142, in <module>
    test(args.secret, 1)
  File "~/tf-encrypted/./test_rss_eq.py", line 128, in test
    result = prot.equal(x, 0)
  File "~/tf-encrypted/tf_encrypted/protocol/protocol.py", line 117, in cache_nodes
    result = func(self, *args, **kwargs)
  File "~/tf-encrypted/tf_encrypted/protocol/aby3/aby3.py", line 1684, in equal
    return self.dispatch("equal", x, y)
  File "~/tf-encrypted/tf_encrypted/protocol/aby3/aby3.py", line 2038, in dispatch
    return func(self, *args, **kwargs)  # pylint: disable=not-callable
  File "~/tf-encrypted/tf_encrypted/protocol/aby3/aby3.py", line 3053, in _equal_private_public
    return _equal_zero_private(prot, x - y)
  File "~/tf-encrypted/tf_encrypted/protocol/aby3/aby3.py", line 3064, in _equal_zero_private
    msb = prot.msb(pack)
  File "~/tf-encrypted/tf_encrypted/protocol/protocol.py", line 117, in cache_nodes
    result = func(self, *args, **kwargs)
[...]

PR #901, opened 2024-09-24
Fix bug: expected always unsigned, but result is signed for some params

zicofish:
Fix bug in this issue #900.

Problem:
`range` is expected to be an `unsigned` type, but for some params (e.g., low=-128, high=126), the usage of `auto` makes `range` becomes a `signed` type. The fix is to explicitly denote `uT` type for `range`.


vincehong (approved):
LGTM. Thanks @GuopengLin for pointing this out.

